PR #6, opened 2026-02-19
allow configuring the chrome profile

pepegar:
Hi!  I use two profiles, one for wokr and one personal, and none of them are called Default, so this extension breaks.

Are you interested in this?  Here I allow configuring the profile that you want to use to get the cookies from.

### DIFF
--- a/chrome-cookies.ts
+++ b/chrome-cookies.ts
@@ -33,14 +33,19 @@ const ALL_COOKIE_NAMES = new Set([
 	"SIDCC",
 ]);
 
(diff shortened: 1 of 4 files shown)


nicobailon:
Thanks @pepegar! The `chromeProfile` config field is in, read from `~/.pi/web-search.json`. I went with runtime path composition via `BrowserConfig.baseDir` + profile name rather than a full path, so it naturally works across all supported browsers — if a browser doesn't have that profile directory, it just falls through to the next one.

Commit: 04dea82

PR #8, opened 2026-02-25
feat: Added Arc Browser support for MacOS

Zeng-Zer:
### Changes

 Add Arc browser support for MacOS:
 - Extended cookie extraction to support both Chrome and Arc browsers
 - Browsers are tried in order: Chrome, Arc

 Refactor:
 - Extracted BrowserConfig interface and BROWSER_CONFIGS array for cleaner browser configuration
 - Added optional browserName parameter to getGoogleCookies()
 - Filtered Browsers based on REQUIRED_COOKIES constant for Gemini auth
 - Updated readKeychainPassword() to accept service name parameter

 ### Before

 - Only Chrome cookies were extracted

 ### After

 - Both Chrome and Arc are tried

### DIFF
--- a/chrome-cookies.ts
+++ b/chrome-cookies.ts
@@ -6,6 +6,9 @@ import { join } from "node:path";
 
 export type CookieMap = Record<string, string>;
 
+// Required cookies for Gemini authentication
+export const REQUIRED_COOKIES = ["__Secure-1PSID", "__Secure-1PSIDTS"];
+
 const GOOGLE_ORIGINS = [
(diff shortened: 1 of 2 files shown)


nicobailon:
Thanks @Zeng-Zer! Arc is supported now as a fallback after Helium and Chrome on macOS. One thing I noticed — your PR dropped the `-a` (account) parameter from the `security find-generic-password` call, which can break on setups where multiple keychain entries share the same service name. The consolidated version keeps both `-a` and `-s`.

Commit: 04dea82

PR #10, opened 2026-02-26
feat: add Linux support for Chrome cookie extraction

anurag-arjun:
## Summary

Extends `getGoogleCookies()` in `chrome-cookies.ts` to work on Linux in addition to macOS, enabling the **Gemini Web search provider** on Linux systems.

## Changes

- **Linux cookie path**: Added `~/.config/google-chrome/Default/Cookies` alongside the existing macOS path
- **Linux password retrieval**: New `readLinuxChromePassword()` that reads from GNOME Keyring via `secret-tool` (tries v2 schema, then v1 schema), falling back to Chrome's default `"peanuts"` password when no keyring is available
- **Platform-appropriate PBKDF2 iterations**: 1003 on macOS (unchanged), 1 on Linux (matches Chromium source)
- **Widened platform gate**: From macOS-only to macOS + Linux

## How it works

On Linux, Chrome encrypts cookies using AES-128-CBC with a key derived via PBKDF2. The encryption password is stored in GNOME Keyring (or KDE Wallet via Secret Service bridge). When no keyring is accessible, Chrome falls back to the hardcoded password `"peanuts"` with 1 PBKDF2 iteration.

## Testing

Tested on CachyOS (Arch-based) with Google Chrome. The standard cookie path and `secret-tool` approach should work across all major Linux distros (Ubuntu, Fedora, Arch, etc.).

## Notes

- Does not cover Chromium, Chrome Beta/Dev, or Flatpak/Snap variants (different cookie paths) — same as the existing macOS code only covers standard Chrome
- No breaking changes to existing macOS behavior

### DIFF
--- a/chrome-cookies.ts
+++ b/chrome-cookies.ts
@@ -33,31 +33,55 @@ const ALL_COOKIE_NAMES = new Set([
 	"SIDCC",
 ]);
 
-const CHROME_COOKIES_PATH = join(
+const CHROME_COOKIES_PATH_DARWIN = join(
 	homedir(),
 	"Library/Application Support/Google/Chrome/Default/Cookies",
 );
 
+const CHROME_COOKIES_PATH_LINUX = join(
+	homedir(),
+	".config/google-chrome/Default/Cookies",
+);
+
+function getChromeCookiesPath(): string {
+	if (platform() === "darwin") return CHROME_COOKIES_PATH_DARWIN;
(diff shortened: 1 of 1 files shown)


anurag-arjun:
Closing as duplicate of #3 which covers the same Linux cookie extraction support (and additionally includes Chromium path support + README updates).

PR #14, opened 2026-03-20
fix: reduce repeated macOS Keychain prompts and auto-detect non-default Chrome profiles

kevinQTruong:
## What changed
- Preflight required Gemini cookie names in SQLite before touching Keychain.
- Auto-scan Chromium profiles when `chromeProfile` is unset.
- Cache browser encryption passwords in-process to avoid repeated Keychain prompts in the same Pi session.

## Verification
- Missing required cookies: `1 security call` -> `0`
- Repeated same-session calls: `2 security calls` -> `1`
- Non-default profile repro: `Profile 2` found automatically
- Real local Chrome DB check still returns the full Gemini auth cookie set

## Included files
- `chrome-cookies.ts`
- `README.md`
- `CHANGELOG.md`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 All notable changes to this project will be documented in this file.
 
+## [Unreleased]
+
+### Fixed
+- Reduced repeated macOS Keychain prompts during Gemini Web cookie lookup by checking for required Google cookie names before reading the browser encryption password.
+- Gemini Web cookie lookup now auto-scans detected Chromium profiles when `chromeProfile` is unset, reducing false negatives for users signed into a non-Default profile.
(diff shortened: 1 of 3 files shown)


josejg:
I tried this, installing with `pi install git:github.com/kevinQTruong/pi-web-access@f13dbb7816445a7f2322aec2e9b0650fad0776c2` and I'm still getting repeated prompts in macOS 26.3.1

luxus:
i tested it as well.. it works for me!

nicobailon:
Thanks for the Keychain/profile work here. I landed a replacement in #157 that replays the useful pieces on current main while preserving the browser-cookie opt-in boundary, adding SQLite backend fallbacks/diagnostics, avoiding cached password-store failures, and blocking profile path/symlink escapes. Closing this PR as superseded by #157.

PR #25, opened 2026-04-08
feat: add native OpenAI web search provider

markoonakic:
## Summary

- add a native `provider: "openai"` path for OpenAI/Codex web search in `pi-web-access`
- make `auto` prefer OpenAI/Codex native search first when the request is compatible
- fall back from `auto` through Exa, then Perplexity, then Gemini
- keep OpenAI capability boundaries explicit: `freshness` supported, allow-list domain filtering only, no `recencyFilter`
- defer Gemini Web probing until fallback actually needs it, avoiding browser cookie/keychain access during normal OpenAI/Exa searches

## Auto provider order

`auto` now uses this order:

1. OpenAI/Codex native web search
2. Exa
3. Perplexity
4. Gemini API
5. Gemini Web

OpenAI is skipped in `auto` mode when the request is incompatible with the native OpenAI provider, specifically when:

- `recencyFilter` is set
- `domainFilter` contains excluded domains (for example `-reddit.com`)

Gemini Web is now treated as a lazy last-resort fallback. We do **not** probe browser cookies/keychain just to build provider availability for normal searches.

## Why this approach

This keeps the feature aligned with Pi's existing auth/model infrastructure instead of requiring the local Codex CLI, while also making `auto` behave closer to user expectations for a real fallback chain. The OpenAI path uses Pi-managed `openai-codex` auth when available and otherwise falls back to standard `openai` API-key auth.

The Gemini Web change fixes noisy browser-storage prompts on macOS by avoiding eager cookie/keychain probing unless Gemini fallback is actually reached.

## Files changed

- `openai-search.ts`
  - native OpenAI/Codex auth resolution
  - Responses API/Codex backend request construction
  - streamed response parsing for `web_search_call` + cited message output
  - OpenAI-specific `freshness` and allow-list `domainFilter` handling
  - rejects unsupported OpenAI `recencyFilter`
- `search-provider-order.ts`
  - compatibility-aware auto ordering with OpenAI-first selection when supported by the request
- `gemini-search.ts`
  - changes `auto` from a single-provider pick into an ordered fallback pass
  - adds OpenAI-first auto handling with runtime fallback through Exa, Perplexity, and Gemini
  - fixes the `isGeminiApiAvailable is not defined` runtime error
  - defers Gemini Web probing until fallback actually reaches Gemini
- `index.ts`
  - makes curator bootstrap/default provider respect OpenAI auto compatibility rules
  - stops generic provider bootstrap from probing Gemini Web just to build availability
  - keeps automatic `web_search` runs on true `auto` so fallback still works in the curated path
  - updates tool help text to match the new ordering
- `chrome-cookies.ts`
  - reorders macOS browser probing to prefer Chrome, Arc, and Brave before Helium
- `README.md`, `CHANGELOG.md`
  - document the OpenAI-first auto ordering and lazy Gemini Web probing
- `test/openai-search.test.ts`, `test/search-provider-order.test.ts`
  - verify OpenAI capability boundaries and compatibility-aware auto ordering

## Notes for reviewers

- explicit `provider: "openai"` remains available and still fails fast on unsupported parameters
- `auto` now includes OpenAI first only when the request shape is compatible
- OpenAI supports:
  - `freshness: "cached" | "live"`
  - allow-list domain filtering only
- OpenAI does not support:
  - `recencyFilter`
  - exclude-domain filters
- Gemini Web is still available as a final fallback, but it is no longer probed during generic availability checks

## Verification

I verified this with:

1. `npm test`
2. a `pi -p` subprocess check using `provider: auto` after the patch (auto no longer crashes)
3. OpenAI web-search docs / Responses API docs review for `external_web_access` and `filters.allowed_domains`
4. root-cause inspection of the Gemini Web cookie/keychain path showing that browser prompts came from eager `isGeminiWebAvailable()` probing during provider bootstrap

## Follow-up ideas (not in this PR)

- add a config switch to disable Gemini Web probing entirely on machines where browser-cookie access is undesirable
- allow explicit browser preference in `web-search.json` for Gemini Web cookie lookup
- consider an `auto-strategy` config knob for users who prefer cheaper-first or provider-specific routing


nicobailon:
Thanks for the native OpenAI search work. Current main now has OpenAI/Codex Responses API web search, provider routing, Codex/OpenAI auth, and tests, so this PR is superseded rather than something we should merge wholesale. I’m closing it to keep the queue accurate. The freshness/domain-filter ideas from this branch are noted for a separate focused follow-up if we adopt them, with credit.

PR #27, opened 2026-04-11
fix: switch code_search to Exa web_search_exa

### LINKED ISSUE #24: code_search breaks because hosted Exa MCP no longer exposes get_code_context_exa by default
`code_search` currently calls Exa MCP tool `get_code_context_exa` against the bare hosted endpoint:

- `code-search.ts` calls `get_code_context_exa`
- `exa.ts` uses `https://mcp.exa.ai/mcp`

In practice this currently fails with:

`MCP error -32602: Tool get_code_context_exa not found`

Reason seems to be that the hosted Exa MCP endpoint no longer exposes `get_code_context_exa` by default, even though older docs suggest it does. I opened an upstream issue here:

https://github.com/exa-labs/exa-mcp-server/issues/275

So `code_search` appears broken unless Exa is called with `?tools=get_code_context_exa` (or the implementation is updated to follow Exa's current recommended tool setup).

dvic:
Fixes #24.

Switch `code_search` off deprecated `get_code_context_exa` and onto `web_search_exa`, per Exa's recommendation in https://github.com/nicobailon/pi-web-access/issues/24#issuecomment-4210440807.

Changes:
- call `web_search_exa` instead of `get_code_context_exa`
- map `maxTokens` to `contextMaxCharacters`
- add a Bun test covering the MCP tool name/args
- update changelog notes

Tested:
- `bun test`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Fixed
+- Switched `code_search` from deprecated Exa MCP `get_code_context_exa` to `web_search_exa`, approximating `maxTokens` with `contextMaxCharacters` so hosted Exa MCP works again without extra tool configuration.
+
 ## [0.10.6] - 2026-04-04
 
 ### Changed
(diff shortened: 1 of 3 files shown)


dvic:
Question: with the switch from `get_code_context_exa` to `web_search_exa`, `code_search.maxTokens` no longer seems to map directly to Exa's token limit. Right now we convert it to `contextMaxCharacters` and cap at 50k chars, so larger requests may silently return much less context than the public `maxTokens` contract implies.

Should we fail fast / lower the documented limit instead of silently truncating, or is there a better way to preserve the existing `maxTokens` behavior with `web_search_exa`?

nicobailon:
Thanks for the patch. The old `code_search` tool has since been removed from main, so this PR is superseded by the shipped removal rather than something we should merge now.

PR #29, opened 2026-04-12
fix: use workflow config instead of hardcoded summary-review

### LINKED ISSUE #28: web_search opens curator browser even when workflow is set to 'none'
**Description**
Setting \`workflow\` to \`"none"\` in \`~/.pi/web-search.json\` does not prevent the search curator from opening in a new browser tab.

**Expected behavior**
When \`workflow\` is set to \`"none"\`, \`web_search\` should return results without opening any browser window.

**Root cause**
In \`index.ts\` line 1147, the \`curatorWorkflow\` variable is hardcoded to \`"summary-review"\` instead of using the resolved \`workflow\` value from config:

\`\`\`typescript
const curatorWorkflow: CuratorWorkflow = "summary-review"; // Bug: hardcoded
\`\`\`

**Fix**
Change line 1147 to:
\`\`\`typescript
const curatorWorkflow: CuratorWorkflow = workflow;
\`\`\`

**Steps to reproduce**
1. Add \`"workflow": "none"\` to \`~/.pi/web-search.json\`
2. Run \`web_search\` with any query
3. Observe: A browser tab still opens with the curator UI

**Environment**
- pi version: 0.37.3+
- pi-web-access version: 0.10.6

sinamtz:
**Summary**
The \`curatorWorkflow\` variable was hardcoded to \`"summary-review"\` instead of using the resolved \`workflow\` value from config, causing the curator to always open regardless of the user's \`workflow\` setting.

**Fix**
Changed line 1147 to use the \`workflow\` variable instead of hardcoding \`"summary-review"\`:

\`\`\`diff
- const curatorWorkflow: CuratorWorkflow = "summary-review";
+ const curatorWorkflow: CuratorWorkflow = workflow;
\`\`\`

**Testing**
1. Set \`"workflow": "none"\` in \`~/.pi/web-search.json\`
2. Run \`web_search\` - no curator opens ✓
3. Set \`"workflow": "summary-review"\` - curator opens as expected ✓

**Related Issue**
Fixes #28

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -1144,7 +1144,7 @@ export default function (pi: ExtensionAPI) {
 				const availableProviders = bootstrap.availableProviders;
 				const defaultProvider = bootstrap.defaultProvider;
 				const curatorTimeoutSeconds = bootstrap.timeoutSeconds;
-				const curatorWorkflow: CuratorWorkflow = "summary-review";
+				const curatorWorkflow: CuratorWorkflow = workflow;
 
 				const summaryContext: SummaryGenerationContext = {
 					model: ctx.model,


nicobailon:
Thanks for the fix. The workflow handling has since been reworked on main: `workflow: "none"` now skips curator mode, and the remaining `summary-review` constant is only inside the already-gated curator path. Closing this as superseded/fixed.

PR #33, opened 2026-04-17
fix(pdf): shim Promise.try on older runtimes

guwidoe:
Hi, I ran into a serious bug where PDF extraction can crash Pi hard.

## Fix

This PR adds a targeted `promise.try` compatibility shim in `pdf-extract.ts`.

Behavior:
- if `Promise.try` already exists, nothing is patched
- if it is missing, `pdf-extract.ts` installs the dedicated `promise.try` shim right before loading `unpdf`

That keeps newer `unpdf` working on older Pi runtimes without forcing everyone onto an older `unpdf` version.

## Why this matters

PDF extraction can currently crash the whole Pi process with:

```text
TypeError: Promise.try is not a function
```

This is a serious bug because when Pi dies mid-session it can leave the terminal in a broken raw/protocol state until the user manually resets it.

## Root cause

Fresh installs can resolve to newer `unpdf` releases whose bundled PDF.js code uses `Promise.try`.

On the Pi runtimes I tested, Node 20 crashes here, and Node 22 has the same issue.

## Why this approach

Pi's current minimum Node requirement is `>=20.6.0`, so I don't think `pi-web-access` can assume all Pi users are already on Node 24+.

I also didn't want to solve this by pinning everyone to an older `unpdf` release. The `promise.try` shim seemed like the smallest fix that stays explicit about what is happening and only changes behavior on runtimes where `Promise.try` is actually missing.

## Revisit later

Once upstream Pi's effective runtime baseline moves to Node 24+, this should be revisited. At that point the shim may no longer be necessary.

## Tests

Added:
- explicit repro for the broken `unpdf@1.6.0` path without the shim
- fixture-backed success test for the current shimmed path
- install-level guard that the package keeps the dedicated shim and modern `unpdf` line

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ All notable changes to this project will be documented in this file.
 ### Fixed
 - Added a targeted `promise.try` compatibility shim in `pdf-extract.ts` so PDF extraction works on Pi runtimes that still lack native `Promise.try()` while leaving newer runtimes untouched.
 - Added regression coverage with a bundled sample PDF fixture, including an explicit repro for the broken `unpdf@1.6.0` path without the shim.
+- Reduced noisy PDF.js warning spam during PDF fetches by forcing the extraction path to run with errors-only PDF.js verbosity.
 
 ## [0.10.6] - 2026-04-04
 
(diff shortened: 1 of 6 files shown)


nicobailon:
Thanks for the fix. I landed a clean current-main replacement in #148 that preserves the Promise.try compatibility behavior without the stale package/test-runner changes, with changelog credit for your PR. Closing this as superseded.

PR #34, opened 2026-04-17
fix(pdf): suppress noisy PDF.js warnings

guwidoe:
Hi, I keep seeing this during PDF fetches in Pi:

warning spam screenshot

The fetch still seems to continue, but the repeated PDF.js warnings spill directly into Pi's interactive UI and make it look broken.

## Proposed fix

Run the PDF extraction path with errors-only PDF.js verbosity:

- import `VerbosityLevel` from `unpdf`
- call `getDocumentProxy(..., { verbosity: VerbosityLevel.ERRORS })`

That should keep real PDF failures visible while suppressing warning spam like:

```text
Warning: CCITTFaxStream: Falling back to JS CCITTFax decoder.
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Fixed
+- Reduced noisy PDF.js warning spam during PDF fetches by running PDF extraction with errors-only PDF.js verbosity.
+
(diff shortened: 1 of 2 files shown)


nicobailon:
Thanks for the fix. I landed the PDF.js errors-only verbosity behavior in #148 alongside the Promise.try compatibility fix, with changelog credit for your PR. Closing this as superseded.

PR #47, opened 2026-04-29
Fix code_search after Exa MCP tool removal

minzique:
## Problem

`code_search` is broken against the current Exa MCP server. The package still calls `get_code_context_exa`, but that tool is no longer advertised by Exa. Right now the server only returns `web_search_exa` and `web_fetch_exa`, so every `code_search` call fails with:

```text
MCP error -32602: Tool get_code_context_exa not found
```

That means Pi users lose code/docs search unless they avoid this tool entirely.

## Fix

This keeps the old path when it works, but handles the removed tool cleanly:

- try `get_code_context_exa` first for backwards compatibility
- if Exa says the tool is missing, cache that and use `web_search_exa` instead
- shape fallback queries toward code, docs, GitHub, Stack Overflow, and official references when the user query does not already do that
- trim the returned text to the requested `maxTokens` budget
- document the fallback in the README and changelog

It is not a perfect replacement for Exa's old code-context tool, but it is much better than returning a hard error. In practice it restores useful code and documentation results with the MCP tools Exa exposes today.

## Verification

- `bun build code-search.ts --target node --outdir /tmp/pi-web-access-check`
- live `executeCodeSearch()` call against the current Exa MCP server returned results with `mode: "web-search-fallback"`

Example query tested:

```text
Evolution API instance connect QR code webhook Baileys GitHub implementation
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Fixed
+- Restored `code_search` after Exa removed the `get_code_context_exa` MCP tool by falling back to `web_search_exa` with a code/documentation-focused query.
+
 ## [0.10.6] - 2026-04-04
 
 ### Changed
--- a/README.md
+++ b/README.md
@@ -99,7 +99,7 @@ web_search({ queries: ["query 1", "query 2"], workflow: "summary-review" })
 
 ### code_search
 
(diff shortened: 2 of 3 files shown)


nicobailon:
Thanks for the fix. I landed a cleaned-up version of this approach in `6705ef8`: keep the old code-context path when available, fall back to `web_search_exa` when missing, add code-focused query shaping, and trim fallback output to the requested token budget. Closing this PR as superseded by the landed commit.

PR #48, opened 2026-04-29
fix(pdf): support Node 22 extraction

### LINKED ISSUE #41: PDF extraction crashes on Node 22 when unpdf calls Promise.try
When pi-web-access fetches a PDF-like URL, it routes the response into the PDF extraction path in extract.ts. In my environment that path crashes with:\n\nTypeError: Promise.try is not a function\n    at file:///tmp/pi-extensions/npm/f35b2129/node_modules/unpdf/dist/pdfjs.mjs\n\nRepro evidence:\n- pi-web-access@0.10.6\n- npm ls pi-web-access --depth=0 -> pi-web-access@0.10.6\n- npm ls unpdf pdfjs-dist pi-web-access --all -> pi-web-access@0.10.6 -> unpdf@1.6.0\n- node -p "process.version + ' Promise.try=' + typeof Promise.try" -> v22.16.0 Promise.try=undefined\n- node_modules/pi-web-access/extract.ts sends PDF responses to extractPDFToMarkdown(buffer, url)\n\nThe issue appears intermittent because only URLs detected as PDFs hit this path; HTML pages do not.\n\nSuggested fix:\n- avoid depending on Promise.try in the bundled PDF path, or\n- add a compatibility guard/polyfill before loading unpdf, or\n- switch to a PDF.js build compatible with the supported Node runtime.\n\nThis is a daily-updated pi extension, so a fix committed only to node_modules would be overwritten on update. The durable fix should live upstream in the repo or in a pinned fork / preload wrapper.

### LINKED ISSUE #46: TypeError: Promise.try is not a function in unpdf pdfjs.mjs when parsing PDF sources
## Environment

- Feynman: 0.2.40
- pi-web-access: 0.10.6
- unpdf (bundled dep): 1.6.0
- Node.js: v22.22.2
- Platform: Raspberry Pi 4 (ARM64), Debian
- Model in use: `anthropic/claude-haiku-4-5-20251001`

## Query that triggered the crash

```
/lit Linux PSI baseline statistical method non-Gaussian distribution
```

## Steps to reproduce

Run `/lit` with a query likely to pull academic PDF sources (statistics, kernel internals). Crash occurs during PDF parsing.

## Error

```
TypeError: Promise.try is not a function
    at #t (file:///.../pi-web-access/node_modules/unpdf/dist/pdfjs.mjs:54:112364)
    at file:///.../pi-web-access/node_modules/unpdf/dist/pdfjs.mjs:54:110186
    at process.processTicksAndRejections (node:internal/process/task_queues:103:5)

Node.js v22.22.2
```

## Notes

`Promise.try` was added in Node.js v22.13.0, which this environment exceeds (v22.22.2). The bundled `pdfjs.mjs` in unpdf 1.6.0 appears to be running in an execution context where the native `Promise.try` is not available. Upgrading unpdf or adding a `Promise.try` polyfill before the PDF parsing call would fix this.

lajarre:
## Summary

- Bump `unpdf` to `^1.6.2`, which includes a `Promise.try` fallback in its bundled PDF.js build.
- Add a Node test that exercises `extractPDFToMarkdown` with an in-memory PDF.
- Run the test in a child process after removing `Promise.try` so the regression covers Node runtimes where `Promise.try` is unavailable.

Fixes #41.
Fixes #46.

## Evidence

| Case | Result |
| --- | --- |
| `unpdf@1.6.0` + Node v22.22.0 | test fails with `TypeError: Promise.try is not a function` from `unpdf/dist/pdfjs.mjs` |
| `unpdf@1.6.2` + Node v22.22.0 | test passes |

## Testing

```bash
npm test
```

```text
# tests 1
# pass 1
# fail 0
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,18 +1,19 @@
 {
   "name": "pi-web-access",
-  "version": "0.7.2",
+  "version": "0.10.6",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "pi-web-access",
-      "version": "0.7.2",
+      "version": "0.10.6",
+      "license": "MIT",
       "dependencies": {
         "@mozilla/readability": "^0.5.0",
         "linkedom": "^0.16.0",
         "p-limit": "^6.1.0",
         "turndown": "^7.2.0",
-        "unpdf": "^1.4.0"
+        "unpdf": "^1.6.2"
       }
     },
     "node_modules/@mixmark-io/domino": {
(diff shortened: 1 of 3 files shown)


championswimmer:
cc: @nicobailon this actually is the correct fix.

PR #49, opened 2026-05-01
Localize web access status and curator strings

jerryfan:
I found one focused UX issue: web access has dense status/error text at the moment users choose search curation, wait for browser approval, fetch URLs, or troubleshoot Gemini/browser setup. If those messages are misread, users can think search is stuck, miss the approval step, or fail provider setup.

This PR makes that path optionally localizable while preserving the current English strings as fallbacks.

- No new dependency
- No behavior change without an i18n provider
- English remains the default/fallback
- Locales: ja, zh-CN, es — broad web/research workflow locales
- Covers curator commands, open/failure notifications, search/fetch progress, and Gemini Web availability

Validation:
- npm install --ignore-scripts
- npm pack --dry-run

### DIFF
--- a/i18n.ts
+++ b/i18n.ts
@@ -1,0 +1,114 @@
+import type { ExtensionAPI } from "@mariozechner/pi-coding-agent";
+
+type Params = Record<string, string | number>;
+type Translate = (key: string, fallback: string, params?: Params) => string;
+
+let translate: Translate = (_key, fallback, params) => format(fallback, params);
+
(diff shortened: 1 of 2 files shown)


jerryfan:
Retracting this PR. I used locale choices that were too broad and not sufficiently supported by explicit repo/product signals. That was not a good fit for a first-touch contribution. I will come back only with a smaller, better-scoped PR using the project’s actual signals and/or a safer Western-default locale set.

PR #50, opened 2026-05-01
Add localized README quickstarts

jerryfan:
This follows the web-access localization direction with docs only.

The package page is where users decide whether web/video/GitHub fetching is safe and worth installing. I translated only the quickstart/decision path, not the full README, so the PR stays reviewable.

- No code changes
- English README unchanged except small language links
- Adds concise ja/zh-CN/es quickstarts
- Focuses on install, provider fallback, GitHub behavior, video support, and optional ffmpeg/yt-dlp setup

I kept commands, config keys, tool names, and code examples unchanged.

### DIFF
--- a/README.es.md
+++ b/README.es.md
@@ -1,0 +1,58 @@
+# Pi Web Access: inicio rápido
+
+[English](./README.md) | [日本語](./README.ja.md) | [简体中文](./README.zh-CN.md)
+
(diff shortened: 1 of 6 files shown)


jerryfan:
Retracting this PR. I used locale choices that were too broad and not sufficiently supported by explicit repo/product signals. That was not a good fit for a first-touch contribution. I will come back only with a smaller, better-scoped PR using the project’s actual signals and/or a safer Western-default locale set.

PR #72, opened 2026-05-19
Fix Gemini Web streamed response parsing

Geometrically:
## Summary
- Scan all Gemini Web streamed response parts instead of stopping at the first candidate list
- Keep the latest non-empty candidate text so YouTube/tool responses do not return empty output
- Preserve existing card-content alternate text handling

## Context
Gemini Web can stream an initial empty candidate before later non-empty chunks, especially for YouTube/tool responses. The previous parser selected the first candidate list and could return an empty response even though the final stream contained valid text.

## Testing
- Verified locally with Gemini Web cookie auth against a YouTube URL; previously failed with `Gemini Web returned empty response`, now returns the expected video/song title.

### DIFF
--- a/gemini-web.ts
+++ b/gemini-web.ts
@@ -366,7 +366,7 @@ function parseStreamGenerateResponse(rawText: string): GeminiWebResult {
 	const errorCode = extractErrorCode(responseJson);
 
 	const parts = Array.isArray(responseJson) ? responseJson : [];
-	let body: unknown = null;
+	let text = "";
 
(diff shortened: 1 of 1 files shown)


nicobailon:
Thanks for the patch. This is the same Gemini stream parsing issue as #54, which I just merged. I went with that version because it keeps the non-empty chunk behavior and preserves the card-content fallback more explicitly, so this PR is now superseded. Appreciate the fix.

PR #76, opened 2026-05-25
Add support for configurable API base URL for gemini, including via Cloudflare AI Gateway

### LINKED ISSUE #74: Support Gemini-compatible gateways — hardcoded API base URL ignores GOOGLE_GEMINI_BASE_URL, conflates host with API version, and contents requests omit role (rejected by Vertex AI passthroughs)
# Support Gemini-compatible gateways — hardcoded API base URL ignores `GOOGLE_GEMINI_BASE_URL`, conflates host with API version, and `contents` requests omit `role` (rejected by Vertex AI passthroughs)

## Problem

`pi-web-access` cannot route Gemini calls (`web_search` with `provider: "gemini"`, URL context extraction, video analysis) through a Gemini-compatible gateway — TrueFoundry, LiteLLM proxy, Helicone, self-hosted Vertex AI passthroughs, regional Google endpoints, etc.

Many organizations require all LLM traffic to flow through a central gateway for billing, audit, rate-limiting, and compliance. The official `gemini` CLI already supports this via `GOOGLE_GEMINI_BASE_URL`, but `pi-web-access` hardcodes the base URL, so users in those environments cannot use `provider: "gemini"` even though they have valid Gemini credentials via their gateway.

There are three underlying gaps. All three need to be addressed for end-to-end gateway routing to work cleanly, and two of them are real concerns even for users who never touch a gateway.

## Current Behavior

1. `API_BASE` is hardcoded in `gemini-api.ts` to `https://generativelanguage.googleapis.com/v1beta` and used directly in all four Gemini call sites (`gemini-search.ts`, `gemini-url-context.ts`, `gemini-api.ts` itself, `video-extract.ts`). There is no env var or config key to override it.
2. Setting `GEMINI_API_KEY=<gateway-token>` and trying `web_search provider:"gemini"` sends the request to Google's endpoint with a gateway-issued token, which Google rejects: `Gemini API error 400: API_KEY_INVALID`.
3. If a user manually patches the base URL to point at a Vertex-backed proxy and retries, the request now reaches the proxy but is rejected with `Gemini API error 400: {"status":"failure","message":"Please use a valid role: user, model.",...}` — because the request body sends `contents: [{ parts: [...] }]` without a `role` field.

## Root Cause Analysis

There are **three** independent gaps:

### Gap 1: `API_BASE` conflates host with API version

```typescript
// gemini-api.ts:5
export const API_BASE = "https://generativelanguage.googleapis.com/v1beta";
```

`/v1beta` is an API version — a request-path concern, not a host/root concern. Treating it as part of the "base URL" has real consequences:

1. **It blocks a clean override.** When `API_BASE` is overridden to a gateway root (Gap 2 below), every call site silently inherits `/v1beta` from whatever the user set, instead of from the plugin. If a user follows the Gemini CLI convention and sets `GOOGLE_GEMINI_BASE_URL="https://my-proxy.com"`, the plugin produces `https://my-proxy.com/models/...:generateContent` with no version segment — which 404s — and the user has to know to manually append `/v1beta` to their env var.
2. **It is inconsistent with how the Gemini CLI defines the same setting.** The CLI's documented example is `export GOOGLE_GEMINI_BASE_URL="https://my-proxy.com"` ([docs](https://www.geminicli.com/docs/reference/configuration)) — a bare host, no version. If `pi-web-access` ships override support but expects `/v1beta` to be part of the env var, the same env var means different things to the two tools.
3. **It couples the plugin to one API version.** Bumping to a future `v1` or `v2beta` requires editing a constant that's typed as a "base URL" rather than the dedicated version field the change actually represents.

This is a real concern *independent of gateway support* — it's a structural smell in the current code, and the right time to fix it is when Gap 2 is added (since changing one without the other forces an awkward UX on override users).

Suggest splitting into two constants:

```typescript
const DEFAULT_API_BASE = "https://generativelanguage.googleapis.com";  // host only
const API_VERSION = "v1beta";                                          // version only
```

and exposing a `getVersionedApiBase()` helper that joins them. `API_BASE` can be kept as a back-compat string alias (computed at module load) so any external import keeps working.

### Gap 2: No way to override the Gemini API base URL

`API_BASE` is a hardcoded module-level constant. The four call sites template it directly into request URLs:

- `gemini-search.ts:205` — `${API_BASE}/models/${model}:generateContent?key=${apiKey}`
- `gemini-url-context.ts:33` — same shape
- `gemini-api.ts:68` (`queryGeminiApiWithVideo`) — same shape
- `video-extract.ts:356,370` — `${API_BASE}/${fileName}?key=${apiKey}` for file lifecycle

There is no env var, config key, or function override. Compare with the official Gemini CLI, which documents [`GOOGLE_GEMINI_BASE_URL`](https://www.geminicli.com/docs/reference/configuration) for exactly this purpose:

> **`GOOGLE_GEMINI_BASE_URL`**: Overrides the default base URL for Gemini API requests (when using `gemini-api-key` authentication). Must be a valid URL.

A user who already has this env var set for Gemini CLI cannot reuse it with `pi-web-access`.

### Gap 3: `contents[]` entries omit the `role` field

All three `:generateContent` callers build a request body without `role` on the `Content` objects:

- `gemini-search.ts:201` — `contents: [{ parts: [{ text: query }] }]`
- `gemini-url-context.ts:29` — `contents: [{ parts: [{ text: EXTRACTION_PROMPT + url }] }]`
- `gemini-api.ts:110-118` (`queryGeminiApiWithVideo`) — `contents: [{ parts: [{ fileData }, { text: prompt }] }]`

The [Gemini API reference](https://ai.google.dev/api/caching#Content) documents `role` as a field on `Content` (valid values: `user`, `model`). Google's public endpoint defaults the role to `user` server-side when omitted, so today's requests succeed against `generativelanguage.googleapis.com` — but Vertex-backed proxies are stricter and reject the request with `400 "Please use a valid role: user, model."`.

This means even with Gap 1 fixed, gateway routing is still broken until requests include `role: "user"` explicitly. It is also better spec compliance to include it.

### Why these are coupled

Gap 1 is a standalone code-quality concern that's worth fixing on its own merits.

Gap 2 and Gap 3 are not user-visible on their own — Gap 2 doesn't matter without somewhere to send the traffic, and Gap 3 doesn't matter on Google's lenient endpoint. They surface together the moment a user tries to use a gateway.

Fixing Gap 2 *without* first fixing Gap 1 forces gateway users to bake `/v1beta` into their env var, breaking parity with the Gemini CLI's documented usage. Conversely, fixing Gap 1 in isolation is harmless to current users (the back-compat alias keeps `API_BASE` looking identical from the outside). For these reasons all three should ship together.

## Proposed Solution

1. **Separate host from API version in `gemini-api.ts`.** Replace `API_BASE` with `DEFAULT_API_BASE` (host only) and `API_VERSION = "v1beta"`. Add a `getVersionedApiBase()` helper that returns `${host}/${API_VERSION}`. Keep `API_BASE` exported as a back-compat string alias.
2. **Add `GOOGLE_GEMINI_BASE_URL` env var + `geminiBaseUrl` config key** to override the host (not the versioned URL). Precedence: env > config > Google's default. Match the Gemini CLI env var name exactly so users with existing setup get the plugin working with zero extra config. Trailing slashes on the override value should be normalized.
3. **Update the four call sites** to use `getVersionedApiBase()` instead of `API_BASE`, so runtime overrides take effect dynamically.
4. **Add `role: "user"`** to every `contents[i]` entry in the three `:generateContent` request bodies.
5. **Document the new env var and config key** in `README.md` alongside `EXA_API_KEY` / `GEMINI_API_KEY` / `PERPLEXITY_API_KEY`.

## Code References

- **Host-vs-version conflation**: `API_BASE` in `gemini-api.ts:5` — `/v1beta` is baked into the same constant as the host
- **Config interface**: `interface GeminiApiConfig` in `gemini-api.ts:9-11` (currently only `geminiApiKey`)
- **Config loader**: `loadConfig()` in `gemini-api.ts:15-30` (reads `~/.pi/web-search.json`)
- **Call sites needing base-URL refactor**:
  - `gemini-search.ts:5` (import) and `gemini-search.ts:205` (fetch URL)
  - `gemini-url-context.ts:2` (import) and `gemini-url-context.ts:33` (fetch URL)
  - `gemini-api.ts:68` (`queryGeminiApiWithVideo` URL)
  - `video-extract.ts:8` (import), `video-extract.ts:356` (poll URL), `video-extract.ts:370` (DELETE URL)
- **Call sites needing `role: "user"`**:
  - `gemini-search.ts:201`
  - `gemini-url-context.ts:29`
  - `gemini-api.ts:110-118` (the `body` in `queryGeminiApiWithVideo`)
- **Out of scope but worth a thought**: `UPLOAD_BASE = "https://generativelanguage.googleapis.com/upload/v1beta"` in `video-extract.ts:13` is a separate Google service host. Gateways typically don't proxy the upload endpoint, so this should likely stay hardcoded unless a user reports needing it overridden.

## Config Interface (current)

```typescript
interface GeminiApiConfig {
  geminiApiKey?: unknown;
}
```

## Config Interface (proposed)

```typescript
interface GeminiApiConfig {
  geminiApiKey?: unknown;
  geminiBaseUrl?: unknown;   // new — overrides API base URL when GOOGLE_GEMINI_BASE_URL env is unset
}
```

## Expected Behavior

A user should be able to:

1. Set `GOOGLE_GEMINI_BASE_URL` (or `geminiBaseUrl` in `~/.pi/web-search.json`) to a Gemini-compatible gateway root and have `web_search provider:"gemini"`, URL context extraction, and video analysis all route through that gateway.
2. Have `provider:"gemini"` work against Vertex-backed proxies without manually editing request bodies.
3. Continue to use the plugin with no env var / no config key set, with byte-identical behavior to today (no regression for current users on Google's direct endpoint).

### Example env setup

```bash
# Matches the official Gemini CLI env var name — see
# https://www.geminicli.com/docs/reference/configuration
export GOOGLE_GEMINI_BASE_URL="https://my-gateway.example.com/api/llm/gemini/proxy"
export GEMINI_API_KEY="<gateway-issued-token>"
```

### Example config

```json
{
  "exaApiKey": "exa-...",
  "perplexityApiKey": "pplx-...",
  "geminiApiKey": "...",
  "geminiBaseUrl": "https://my-gateway.example.com/api/llm/gemini/proxy",
  "summaryModel": "anthropic/claude-haiku-4-5"
}
```

## Design Notes

A few decisions worth surfacing for review before a PR:

- **Env var name**: `GOOGLE_GEMINI_BASE_URL`, not `GEMINI_BASE_URL` or a plugin-specific prefix — to match the [official Gemini CLI](https://www.geminicli.com/docs/reference/configuration) exactly. Existing Gemini CLI users get this for free.
- **Resolution is a function, not a top-level `const`.** Reading env at call time means runtime overrides work as expected (e.g., tests, dynamic config).
- **Adding `role` everywhere, not just where it broke**: consistency, and it's documented as a real field on `Content`. The public endpoint defaulting it is implementation detail that could change.

## Additional Context

- Verified end-to-end against a corporate TrueFoundry gateway (`https://<tenant>.truefoundry.cloud/api/llm/gemini/google-vertexai/proxy`) using `gemini-3-flash-preview` (the plugin's `DEFAULT_MODEL`). With all three gaps fixed, `web_search provider:"gemini"` returns real Gemini-grounded results through the gateway, identical in shape to results from Google's direct endpoint.
- The official Gemini CLI works against this same proxy URL unmodified with the same env var name, confirming the proxy speaks the Gemini-native API shape — the plugin is the only blocker.
- Affected version: `pi-web-access@0.10.7`.
- The three gaps can be addressed together or split (Gap 1 as a pure refactor, Gaps 2 and 3 together as the gateway-support change). They share a test setup, but Gap 1 is logically independent and a strict no-op for current users — it could land first as a low-risk cleanup.

meatballhat-cf:
I hit the same issue as described in #74 and I work in a corporate environment where I need to specifically use Cloudflare AI Gateway. This is my attempt to accommodate the general case and also ensure Cloudflare AI Gateway works similarly to the support that was recently added to `pi` core.

## Summary

- Split `API_BASE` into `DEFAULT_API_HOST` + `API_VERSION`; add `getVersionedApiBase()`. `API_BASE` kept as deprecated export for backward compatibility.
- Add `GOOGLE_GEMINI_BASE_URL` env var and `geminiBaseUrl` config key to override the Gemini API host (matches the [official Gemini CLI](https://www.geminicli.com/docs/reference/configuration)).
- When the host is a Cloudflare AI Gateway endpoint, automatically switch to `cf-aig-authorization` header auth using `CLOUDFLARE_API_KEY` / `cloudflareApiKey`.
- Add `role: "user"` to all `contents[]` entries for Vertex AI proxy compatibility.
- Update all four call sites to use `getVersionedApiBase()`.

## Config

```json
{
  "geminiBaseUrl": "https://my-gateway.example.com/gemini",
  "cloudflareApiKey": "..."
}
```

Or via env: `GOOGLE_GEMINI_BASE_URL` and `CLOUDFLARE_API_KEY`.

---

Closes #74


nicobailon:
Thanks for the PR. The Gemini base URL and Cloudflare AI Gateway support looks useful, but the branch is stale and conflicts with current `main` in all touched files. Could you rebase on the latest `main` and resolve the conflicts? Main has since changed import specifiers, config-path handling, and search provider routing, so the update will need to fit those current patterns. Once it’s rebased, I’d be happy to review the implementation again.

nicobailon:
Landed this on main as 7cba8ef with the current config/docs/tests and credit. Closing the stale PR branch now.

PR #88, opened 2026-06-04
Normalize fetch_content optional parameters

WaitHZ:
## Summary

Fix `fetch_content` parameter normalization so bridge-filled empty optional fields do not override valid inputs.

This handles cases where a tool bridge/model emits payloads such as:

```json
{
  "url": "https://example.com/docs",
  "urls": [],
  "frames": 1,
  "prompt": "",
  "timestamp": "",
  "model": ""
}
```

Previously:
- `urls: []` took precedence over `url`, resulting in `Error: No URL provided.`
- bridge-filled `frames: 1` caused ordinary webpage fetches to enter frame extraction and fail with `Frame extraction only works with YouTube and local video files`

## Changes

- Add `normalizeFetchContentParams()` helper
- Combine non-empty `urls` and `url` values instead of letting an empty `urls` array suppress `url`
- Trim/drop blank optional strings
- Ignore likely bridge-filled `frames: 1` for ordinary page fetches while preserving explicit video frame cases (`timestamp`, `prompt`, or `frames > 1`)
- Add regression tests for the normalization behavior

## Test plan

```bash
npm test
```

All tests pass locally.

### DIFF
--- a/fetch-params.ts
+++ b/fetch-params.ts
@@ -1,0 +1,67 @@
+export interface FetchContentParams {
+	url?: unknown;
+	urls?: unknown;
+	forceClone?: unknown;
+	prompt?: unknown;
+	timestamp?: unknown;
+	frames?: unknown;
+	model?: unknown;
+}
+
+export interface NormalizedFetchContentParams {
+	urlList: string[];
+	options: {
+		forceClone?: boolean;
+		prompt?: string;
+		timestamp?: string;
+		frames?: number;
+		model?: string;
+	};
+}
+
+export function normalizeFetchContentParams(params: FetchContentParams): NormalizedFetchContentParams {
+	const normalizedUrls = uniqueUrls(normalizeUrlArray(params.urls));
+	const urlList = normalizedUrls.length > 0 ? normalizedUrls : normalizeSingleUrl(params.url);
+	const prompt = normalizeOptionalString(params.prompt);
+	const timestamp = normalizeOptionalString(params.timestamp);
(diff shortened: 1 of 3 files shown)


WaitHZ:
Thanks for the review. I updated the frames heuristic so a non-empty prompt no longer preserves a bridge-filled `frames: 1`; frames are only forwarded when a timestamp is present or frames > 1. I also kept the extra normalization coverage added in the prior update.\n\nRegarding the `.mjs` test importing `.ts`: this repo's existing tests already import TypeScript modules directly (for example `test/pdf-extract.test.mjs` imports `../pdf-extract.ts`), and `npm test` passes locally with the current setup. Importing `../fetch-params.js` would fail in this source-only package because no compiled JS artifact exists in the repo.\n\nLatest local test run: `npm test` → 8 passed, 0 failed.

WaitHZ:
Addressed the latest actionable comments:\n\n- Preserved the previous precedence behavior: non-empty `urls` now overrides `url`; `url` is only used as a fallback when normalized `urls` is empty.\n- Tightened `frames` normalization to accept only positive integers (`>= 1`).\n- Added regression tests for both behaviors.\n\nLatest local test run: `npm test` → 9 passed, 0 failed.\n\nThe remaining `.mjs` → `.ts` import comment is intentionally left as-is because this source-only repo already uses that pattern in existing tests (e.g. `test/pdf-extract.test.mjs` imports `../pdf-extract.ts`), and importing `.js` would fail without compiled artifacts.

PR #94, opened 2026-06-18
fix: add missing typebox runtime dependency

GautamKumarOffical:
The extension imports `Type` from `typebox` at runtime but `typebox` was not declared in `package.json` dependencies, causing clean installs to fail with `Cannot find module 'typebox'`.

Fixes #1828


nicobailon:
Thanks for catching the packaging issue. Current `main` now handles this through Pi-provided `typebox` as a peer dependency, and this PR adds `typebox@^0.34.0`, which does not exist on npm and would break clean installs. I’m going to close this one rather than ask for a rework, but the underlying issue should be covered now.

PR #102, opened 2026-06-25
feat(ssrf): add `ssrf.allowRanges` to exempt CIDR ranges from the SSRF guard

### LINKED ISSUE #101: SSRF guard blocks all fetches under TUN/fake-IP proxy DNS (198.18.0.0/15) — add opt-in toggle
## Summary

`fetch_content` / `web_search` URL fetching is unusable for many users whose machines run a network proxy in **TUN mode with fake-IP DNS** (Surge, Clash, Mihomo, Stash, Loon, Quantumult X, etc. — very common in China, especially for users who need a proxy to reach sites like GitHub). The SSRF guard in `ssrf-protection.ts` flags these requests as "internal" and blocks them, even though the target host is fully public.

There is currently **no config option or env var to relax this** — the block list is hardcoded and `validateRemoteUrl()` is called with no options at every call site. Requesting an opt-in toggle.

## Reproduce

On a machine whose proxy is in TUN + fake-IP mode, `github.com` resolves to a synthetic address inside the reserved `198.18.0.0/15` (RFC 2544) range:

```sh
$ host github.com
github.com has address 198.18.0.56
$ host raw.githubusercontent.com
raw.githubusercontent.com has address 198.18.0.154
```

Then any fetch fails:

```
Error: Blocked internal address for github.com: 198.18.0.56
```

## What is TUN mode + fake-IP, and why it collides here

**TUN mode** (used by Surge/Clash/Mihomo and friends) creates a virtual network interface and captures all traffic at the IP layer, then transparently routes it through the proxy. It's almost always paired with **fake-IP DNS**: instead of returning the real resolved address, the proxy's DNS hands out synthetic placeholder IPs drawn from a reserved range (the de-facto standard is `198.18.0.0/15`). The proxy then matches the *hostname* (kept in its own fake-IP → domain map) to the real upstream when the connection comes in over the TUN interface.

Net effect: to anything that does its own `getaddrinfo` and inspects the result, *every* proxied hostname appears to live at `198.18.x.x`. Browsers work because their traffic is captured at the TUN layer; anything that validates the **resolved IP** before dialing (like this package's SSRF check) sees a "reserved/internal" address and aborts.

The guard is doing exactly what it should:

```ts
// ssrf-protection.ts -> isBlockedIPv4()
(a === 198 && (b === 18 || b === 19)) ||   // ← 198.18.0.0/15 (RFC 2544)
```

That range is correctly reserved, so `isBlockedIPv4` is *right* to distrust it in the general case. The problem is purely environmental — the host is public, the local DNS is lying — and right now there's no way to tell the tool that.

## Impact

- Every `fetch_content` to a proxied domain fails with a hard error, with no workaround on the tool side.
- `web_search` result fetching and the GitHub extractor are similarly affected (they share `extract.ts` → `validateRemoteUrl`).
- The only current remedies are all *outside* the tool: reconfigure the proxy's fake-IP filter to hand out real IPs for specific domains, switch to `redir-host` mode, or shell out to `curl`/`python urllib` (which honor the proxy's transparent routing and bypass this guard entirely).
- Confirmed there is no existing escape hatch: no key in `~/.pi/web-search.json`, no `PI_*` env var, and `validateRemoteUrl` is invoked with default options at `extract.ts:86`, `extract.ts:373`, and `extract.ts:507`.

## Proposal

Add an opt-in way to relax/override the guard. 

 **Allow specific reserved ranges**
   ```jsonc
   "ssrf": { "allowRanges": ["198.18.0.0/15"] }
   ```
   Let the user opt specific ranges out of `isBlockedIPv4`/`isBlockedIPv6`. Good for fake-IP proxies generically.

The setting should default to current behavior (guard fully on) so nothing changes unless explicitly opted in.

### Implementation note

`validateRemoteUrl` and `fetchRemoteUrl` already accept an `options` argument and the blocking logic is centralized in `isBlockedIPv4` / `isBlockedIPv6`, so threading a config-derived allowlist down to those two functions looks like a small, contained change. Happy to open a PR for any of the shapes above

## Related

- #79 — related but orthogonal: that asks for a *user-facing domain allow/deny list* (a permission/UX layer over which hosts the agent may fetch at all). This issue is about the *security guard producing false positives* for genuinely public hosts under fake-IP/TUN proxy DNS. The two compose nicely: #79 decides *whether* to fetch, this decides *whether the SSRF preflight is appropriate* for a given environment.

## Environment

- pi-web-access: `0.12.0`
- OS: macOS (Apple Silicon)
- Proxy: TUN mode + fake-IP DNS (`github.com` → `198.18.0.56`)

TianZuo555:
## Summary

Resolves #101.

`fetch_content` / `web_search` are unusable for users whose machines run a network proxy in **TUN mode with fake-IP DNS** (Surge, Clash, Mihomo, Stash, Loon, Quantumult X, … — very common in China for reaching GitHub etc.). Such proxies resolve public hostnames into a synthetic reserved range (`198.18.0.0/15`), and the SSRF guard correctly flags these as internal and blocks the request — with **no config option to relax it** today.

This PR adds an opt-in `ssrf.allowRanges` config that exempts specific CIDR ranges from `isBlockedIPv4` / `assertPublicAddress`. **Off by default** — the guard stays fully enabled unless ranges are listed.

```jsonc
{
  "ssrf": { "allowRanges": ["198.18.0.0/15"] }
}
```

## What changed

- **`ssrf-protection.ts`**
  - `ValidationOptions` gains `allowRanges?: string[]`.
  - New `parseAllowRanges` / `parseCidr` / `isInAllowedRange` / `bytesMatchPrefix` helpers parse CIDRs into byte arrays + prefix length and do bit-level matching. Same-family comparison only (4-byte IPv4 rules never match 16-byte IPv6 addresses).
  - `assertPublicAddress` exempts explicitly-allowed ranges *before* the private/reserved checks.
  - **Threaded through redirects:** `fetchRemoteUrl` re-validates each hop with the same `options` (incl. `allowRanges`).
- **`extract.ts`**
  - `loadSsrfAllowRanges()` reads `ssrf.allowRanges` from `web-search.json`; returns `[]` (guard fully on) on missing/unreadable/invalid-JSON/unset, and **throws on a non-array value** so a mistyped config fails loudly.
  - All three call sites pass `allowRanges` through; an `options.lookup` test seam is threaded down so SSRF validation can be exercised without real DNS.
- **`index.ts`** — `ssrf?: { allowRanges?: string[] }` added to `WebSearchConfig`.
- **`README.md` / `CHANGELOG.md`** — documented, off-by-default, "use the narrowest range".

## Safety details

- **No `/0` footgun:** `parseCidr` rejects a trailing-slash / non-numeric prefix via `/^\d+$/` rather than letting `Number("")`/`Number(" ")` silently become `/0` (which would exempt *every* address). Explicit test for this.
- **Strict, fail-loud validation:** non-array and malformed CIDR entries throw instead of being silently ignored.
- **No regression for existing users:** empty list short-circuits in `isInAllowedRange`; behavior is identical unless ranges are configured.

## Tests

All 67 tests pass (`node --test test/*.test.mjs`). New coverage:
- Fake-IP range (`198.18.0.0/15`) exemption + IPv6 (`fd00::/8`).
- Non-relaxation outside the listed range; exact `/31` boundary.
- Bare-host `/32` handling; `/0`-via-empty-prefix rejection.
- Descriptive errors for invalid entries; non-array rejection.
- `allowRanges` flows through `fetchRemoteUrl` redirect targets.
- `loadSsrfAllowRanges` fail-safe: missing config, unset key, invalid JSON, non-array; trim/dedupe empty entries.


TianZuo555:
@nicobailon Plz let me know your thoughts on this change, thanks for your time!

PR #106, opened 2026-06-26
feat(olostep): add Olostep as optional search and content extraction provider

Olostep-devs:
## What this adds

Adds Olostep as a fully optional, key-gated provider for both `web_search` and `fetch_content`. Users without an Olostep key see no change in behavior.

### Search (`web_search`)
- New `olostep.ts` provider module uses the [Olostep answers API](https://docs.olostep.com) to return a synthesized answer with source citations in the same shape as other providers
- `provider: 'olostep'` explicit routing added to `gemini-search.ts`
- In `auto` mode, Olostep joins the fallback chain after Tavily when a key is present
- Zero-config Exa default path is **unchanged** — no key means Olostep is silently skipped

### Content extraction (`fetch_content`)
- Olostep scrape (single URL to clean markdown) added as a fallback in `extract.ts` after Jina Reader, for blocked, JS-rendered, or anti-bot pages

### Configuration
Add to `~/.pi/web-search.json`:
```json
{ "olostepApiKey": "olostep-..." }
```
Or set the `OLOSTEP_API_KEY` env var (takes precedence).

### Files changed
- `olostep.ts` — new provider module (search + scrape, modeled on `exa.ts`)
- `gemini-search.ts` — `olostep` added to `SearchProvider` enum, routing, and auto fallback chain
- `extract.ts` — Olostep scrape slotted after Jina Reader in blocked-page retry path
- `index.ts` — provider availability check + guard
- `test/olostep.test.mjs` — 7 tests (all pass)
- `README.md`, `CHANGELOG.md` — updated

### Tests
```
node --test test/olostep.test.mjs
✔ isOlostepAvailable returns false with no key
✔ isOlostepAvailable returns true when OLOSTEP_API_KEY is set
✔ searchWithOlostep returns null with no API key
✔ searchWithOlostep sends correct request and returns shaped result
✔ auto search fallback chain skips Olostep when no key
✔ extractWithOlostep returns null with no API key
✔ extractWithOlostep returns ExtractedContent on success
7/7 pass
```

cc @nicobailon

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Added
+- Added Olostep as a fully optional, key-gated search and content extraction provider. Set `OLOSTEP_API_KEY` or `olostepApiKey` in `~/.pi/web-search.json` to activate. When keyed, `web_search({ provider: "olostep" })` uses Olostep answers/search to return a synthesized answer with source citations. In `auto` mode, Olostep participates in the fallback chain after Tavily when a key is present. `fetch_content` uses Olostep scrape as a fallback for blocked, JS-rendered, or anti-bot pages, slotted after Jina Reader. Users without an Olostep key see no change in behavior.
+
 ## [0.13.0] - 2026-06-25
 
 ### Added
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ https://github.com/user-attachments/assets/cac6a17a-1eeb-4dde-9818-cdf85d8ea98f
 
 **Video Understanding** — Point it at a YouTube video or local screen recording and ask questions about what's on screen. Full transcripts, visual descriptions, and frame extraction at exact timestamps.
 
-**Smart Fallbacks** — Every capability has a fallback chain. Search tries OpenAI when suitable and available, then Exa, Brave, Parallel, Tavily, Perplexity, Gemini API, and Gemini Web when browser cookies are enabled. YouTube tries Gemini Web when enabled, then API, then Perplexity. Blocked pages retry through Jina Reader, Parallel, and Gemini extraction. Something always works.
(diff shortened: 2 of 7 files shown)


Olostep-devs:
Heads up: please hold on merging this. A live test against the Olostep API surfaced two shape bugs that the mocked tests missed: the search call must POST `task` (not `query`) to /v1/answers, and both the answers and scrape responses are read at the wrong nesting (answer is in `result.json_content`, sources in `result.sources`, scrape markdown in `result.markdown_content`). Fixing and re-verifying against the live API before this is ready to merge.

Olostep-devs:
Fixed and verified against the live Olostep API. Search now POSTs `task` and reads the answer from `result.json_content` with citations from `result.sources`; extract reads `result.markdown_content`. Live smoke: search returned a synthesized answer, and a scrape of a JS page returned clean markdown. Tests updated to the real response shapes (7/7 pass). Ready for review.

nicobailon:
Thanks for the Olostep provider and for following up with live API-shape fixes. I reviewed head `932a42a73693401df2d42b20534a270bd7224ac3` against current main. I’m not going to merge it as-is because it predates current credential-source and provider/security wiring.

The main required change is to port credential handling to `credential-source.ts`, including typed credential failures and the current `$ENV`/`!command` config behavior. The rebase also needs to preserve the current SSRF/proxy flow for `fetch_content`, add a test that blocked private targets do not invoke Olostep, and wire Olostep consistently through provider availability/default/schema paths.

The live API-shape corrections are useful evidence. If this is rebased or clean-ported, please keep the focused provider tests and add the current-main credential/security coverage; credit for @Olostep-devs / #106 will be preserved.

nicobailon:
Thanks for the Olostep implementation and for correcting the live API shape in follow-up commits. The current-main pass found that the scrape fallback is the materially different capability here and is worth evaluating, but the PR needs rework before it is safe to consider.

Please rebase onto current main, port credentials to the shared request-time credential-source path, and make sure the extraction path preserves the current SSRF boundary. In particular, add a test proving a blocked/private target never invokes Olostep. It would also help to split or clearly isolate scrape from search so we can review the JS/blocked-page extraction value separately from adding another hosted search provider. Billing/limits and the live error shapes for both endpoints should be documented or captured in tests.

nicobailon:
Thanks for the Olostep work here. I’m closing this PR because the search slice now overlaps too much with the current provider set, and the branch predates the current credential-source, SSRF, Firecrawl, and routing changes.

The extraction idea is still the salvageable part. If we bring it back, it should be a fresh extraction-only change on current main, using the shared credential-source path, preserving the current SSRF boundary, proving blocked private targets make zero Olostep calls, and documenting the third-party URL/billing behavior. If that lands, I’ll credit Olostep Devs / @ZeeshanAdilButt and PR #106 in the changelog.

Issue #105, opened 2026-06-25
Support SearXNG and Firecrawl as providers


avinashkanaujiya:
### Problem Statement
Users running self-hosted SearXNG (metasearch) and Firecrawl (web scraping/crawling) instances can't use them as providers in pi-web-access. These are popular self-hosted alternatives for privacy-conscious setups. Currently the only supported search providers are SaaS: OpenAI, Exa, Brave, Parallel, Tavily, Perplexity, and Gemini.

### Proposed Solution

1. **SearXNG as a `web_search` provider** — SearXNG exposes a standard JSON API (`/search?format=json&q=...`). Add it as a provider with a configurable `baseUrl`, so self-hosted instances "just work."

2. **Firecrawl as a `fetch_content` extraction method** — Firecrawl exposes a scrape endpoint (`/v2/scrape`) that returns clean markdown. Add it as an extraction fallback (alongside Readability, Jina, Gemini URL Context) with a configurable `baseUrl` and optional API key.

Both should support custom base URLs via `~/.pi/web-search.json`:
```json
{
  "searxngBaseUrl": "https://search.example.com",
  "firecrawlBaseUrl": "https://crawl.example.com",
  "firecrawlApiKey": "fc-..."
}
```

### Use Case
- Privacy: no data leaves the user's infrastructure
- Cost: no per-query SaaS billing for high-volume use

### Current Environment / Context
Self-hosted homelab (Docker Compose), Debian Linux. SearXNG and Firecrawl already running as containers locally.

### Would you be willing to implement this?
- [ ] Yes, I would implement this
- [ ] Yes, I could help implement this
- [x] No, I just want this feature

nicobailon:
Thanks for the self-hosted SearXNG/Firecrawl request. I reviewed the current source plus PRs #107 and #123. Both are useful references, but neither should be merged directly on current main.

The issue is actionable after one product decision: if a user configures self-hosted SearXNG or Firecrawl, should `auto` prefer those trusted self-hosted endpoints before hosted providers, or should they remain explicit/configured-provider only? My default recommendation is to treat a configured self-hosted provider as an intentional trust/privacy signal, but that needs to be documented and tested.

If we implement this, it should be a clean current-main replacement that credits @avinashkanaujiya for #105, @marnunez for #107, and @fank for #123, preserves SSRF protections for requested page URLs, and resolves Firecrawl `/v1` vs `/v2` behavior before landing.

PR #107, opened 2026-06-26
feat: add SearXNG search provider

marnunez:
## Summary

Adds SearXNG as an optional `web_search` provider for self-hosted/private metasearch setups. This addresses the SearXNG portion of #105; Firecrawl is intentionally left out to keep this PR focused.

## Details

- Adds `searxngBaseUrl` config and `SEARXNG_BASE_URL` env support
- Uses SearXNG JSON API at `/search?format=json&q=...`
- Supports `numResults`, `recencyFilter`, and domain include/exclude filters
- Wires SearXNG into explicit provider selection, auto fallback, curator availability/buttons, and tool schema
- Updates README/CHANGELOG and provider tests

## Validation

- `npm test`
- Manual smoke test against a self-hosted SearXNG instance (`SEARXNG_BASE_URL=... node --input-type=module ...`)

Refs #105

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+- Added SearXNG as an optional `web_search` provider via `searxngBaseUrl` / `SEARXNG_BASE_URL`, using the self-hosted JSON API and supporting result count, recency, and domain filters.
+
 ## [0.13.0] - 2026-06-25
 
 ### Added
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 
 # Pi Web Access
 
-**Web search, content extraction, and video understanding for Pi agent. OpenAI/Codex search, zero-config Exa search, Brave, Parallel, Tavily, optional browser-cookie Gemini Web, or bring your own API keys.**
+**Web search, content extraction, and video understanding for Pi agent. OpenAI/Codex search, zero-config Exa search, Brave, Parallel, Tavily, SearXNG, optional browser-cookie Gemini Web, or bring your own API keys.**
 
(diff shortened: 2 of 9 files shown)


marnunez:
One design question before this gets reviewed: should a configured SearXNG instance be preferred earlier in `auto` provider selection?

Right now this PR wires SearXNG into auto fallback after Exa/Brave/Parallel/Tavily. Since `isExaAvailable()` is effectively always true because of Exa MCP, a user who sets `searxngBaseUrl` will usually only get SearXNG if they explicitly set `provider: "searxng"`.

Given the privacy/cost motivation in #105, would you prefer `auto` to treat `searxngBaseUrl` as a strong signal and try SearXNG before SaaS/hosted providers, or should it remain explicit-only unless selected as the provider?

nicobailon:
Thanks for the SearXNG provider. I reviewed head `9cd2d1caf1486012f506fd5bfa5b5713251f27db` against current main and issue #105. The provider shape is viable, but this exact branch is stale and should not be merged directly.

The unresolved question in your comment is the important one: if a user configures a self-hosted SearXNG instance, should `auto` prefer it before hosted providers, or should it remain explicit/configured-provider only? That is a product/privacy policy decision and needs to be documented/tested either way.

If this moves forward, it should be clean-ported onto current main, preserve current provider-precedence behavior intentionally, document the trust boundary for a configured self-hosted base URL, and add current-main integration tests. We’ll preserve credit for @marnunez and link this with #105 if the SearXNG slice lands.

nicobailon:
Thanks again for the SearXNG implementation. I landed a current-main replacement in #164 that preserves the core self-hosted SearXNG behavior from this PR, updates it for the current provider/curator surfaces, makes configured SearXNG local-first in auto, and keeps the configured base URL plus redirects under the shared SSRF guard.

I credited Marcos A. Núñez (@marnunez) for #107 in the changelog. Issue #105 remains open for the Firecrawl extraction half.

PR #109, opened 2026-06-28
feat(ssrf): add trust env proxy

DBPhoenix:
Adds an opt-in `ssrf.trustEnvProxy` setting for proxied environments.

When enabled, `pi-web-access` skips local DNS preflight for hostnames sent through explicit `HTTP_PROXY` / `HTTPS_PROXY`, while still blocking `localhost`, `*.localhost`, and literal internal IPs, and respecting `NO_PROXY`.

This is mandatory in sandboxed setups like OpenShell where outbound access is enforced by the sandbox/proxy policy, and local DNS preflight fails before the request reaches that policy boundary.

### DIFF
--- a/extract.ts
+++ b/extract.ts
@@ -21,37 +21,51 @@ const NON_RECOVERABLE_ERRORS = ["Unsupported content type", "Response too large"
 const MIN_USEFUL_CONTENT = 500;
 const WEB_SEARCH_CONFIG_PATH = getWebSearchConfigPath();
 
+interface SsrfConfig {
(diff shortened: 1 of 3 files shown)


nicobailon:
Thanks for the proxy-aware SSRF work. I landed a replacement in #153 that keeps the same opt-in direction, adds validation for proxy env values, documents the boundary that this is DNS preflight only rather than full fetch proxy transport, and includes regression tests. Closing this PR as superseded by #153.

PR #118, opened 2026-07-05
fix: drop pi-ai/compat subpath, use bare package name (jiti subpath alias bug)

chronoAP:
## Problem

`pi-web-access` 0.13.0 imports from the `@earendil-works/pi-ai/compat` subpath in three files. All fail under the current `pi-coding-agent` + `jiti` 2.7.0 combo:

```
Failed to load extension: Cannot find module
`.../pi-ai/dist/index.js/compat`
```

The extension is completely unloadable.

## Root cause

`pi-coding-agent`'s extension loader (`dist/core/extensions/loader.js`) registers jiti aliases mapping `@earendil-works/*` packages to the right entrypoints:

```js
"@earendil-works/pi-ai": piAiCompatEntry,
"@earendil-works/pi-ai/compat": piAiCompatEntry,
```

**jiti 2.7.0 only honors the bare package-name alias.** Subpath aliases containing `/` are ignored: jiti resolves the package main (`dist/index.js`) and appends the subpath as a literal path, producing `dist/index.js/compat` which does not exist.

The loader comment spells out the intended convention:

> Extensions resolve the pi-ai root to the compat entrypoint (a strict superset of the core entrypoint).

## Fix

Drop the `/compat` subpath everywhere. Import from the bare `@earendil-works/pi-ai`, which the loader alias resolves to the compat entrypoint:

```diff
- import { ... } from "@earendil-works/pi-ai/compat";
+ import { ... } from "@earendil-works/pi-ai";
```

Three files changed (all the call sites):

- `index.ts:4` — static import of `StringEnum`, `complete`, `Model`
- `summary-review.ts:1` — static import of `complete`, `Message`, `Model`
- `openai-search.ts:120` — dynamic `await import(...)` for `getModel`

## Verification

Patched the installed extension, fully reloaded pi:

- Before: `[Extension issues] pi-web-access Failed to load extension: Cannot find module ...`
- After: `[Extensions] ... pi-web-access` loads cleanly, no errors.

`StringEnum`, `complete`, `Message`, `Model`, and `getModel` all resolve via the bare-name alias.

## Notes

- The deeper jiti bug (subpath aliases containing `/` not matched) affects any extension using a subpath import. A jiti-side fix would make this more robust, but this change aligns `pi-web-access` with pi documented extension convention regardless.
- `compat` is explicitly a temporary entrypoint per `pi-ai`, so these imports are already on a deprecation path; this change does not increase exposure.

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -1,7 +1,7 @@
 import type { ExtensionAPI, ExtensionContext } from "@earendil-works/pi-coding-agent";
 import { Box, Text, truncateToWidth } from "@earendil-works/pi-tui";
 import { Type } from "typebox";
-import { StringEnum, complete, type Model } from "@earendil-works/pi-ai/compat";
+import { StringEnum, complete, type Model } from "@earendil-works/pi-ai";
 import { fetchAllContent, type ExtractedContent } from "./extract.ts";
 import { normalizeFetchContentParams } from "./fetch-params.ts";
 import { clearCloneCache } from "./github-extract.ts";
--- a/openai-search.ts
+++ b/openai-search.ts
@@ -117,7 +117,7 @@ function extractAccountId(token: string): string | undefined {
 
 export async function resolveOpenAIAuth(ctx?: ExtensionContext): Promise<OpenAIAuth | undefined> {
 	if (ctx) {
-		const { getModel } = await import("@earendil-works/pi-ai/compat");
+		const { getModel } = await import("@earendil-works/pi-ai");
 		for (const candidate of AUTH_MODEL_CANDIDATES) {
 			for (const modelId of candidate.models) {
 				const model = getModel(candidate.provider, modelId);
--- a/summary-review.ts
+++ b/summary-review.ts
@@ -1,4 +1,4 @@
-import { complete, type Message, type Model } from "@earendil-works/pi-ai/compat";
+import { complete, type Message, type Model } from "@earendil-works/pi-ai";
 import type { ExtensionContext } from "@earendil-works/pi-coding-agent";
 import { loadEnabledModelPatterns, modelMatchesEnabledPatterns } from "./summary-model-scope.ts";
 import type { QueryResultData } from "./storage.ts";


nathanpt:
Thanks for the detailed writeup. I tried to reproduce this against a current install and couldn't — I'd like to share what I found before this merges.

## Reproduction attempt

Environment matching the PR's setup:
- `pi-coding-agent@0.80.2`
- `jiti@2.7.0` (confirmed in the host's `node_modules`)

My fork (`nathanpt/pi-web-access`) imports from `@earendil-works/pi-ai/compat` in the same three sites this PR touches (`index.ts`, `curator/summary-review.ts`, `providers/openai-search.ts` dynamic import). Loaded it with `pi -e .`:

```
Error: Failed to load extension ".../pi-web-access/index.ts": Tool "web_search" conflicts with ...
```

That's a **tool-registration conflict** against a second installed copy — and tool registration only happens *after* a module has loaded successfully. If `/compat` had failed to resolve, I'd see the reporter's `Cannot find module '.../pi-ai/dist/index.js/compat'` instead. The `/compat` subpath resolves fine in the current host.

## The loader handles `/compat` in both modes

`pi-coding-agent`'s `dist/core/extensions/loader.js` registers the subpath alias in **both** runtime paths:

```js
// Node/jiti mode (getAliases())
"@earendil-works/pi-ai": piAiCompatEntry,
"@earendil-works/pi-ai/compat": piAiCompatEntry,   // line 87

// Bun binary mode (VIRTUAL_MODULES)
"@earendil-works/pi-ai": _bundledPiAiCompat,
"@earendil-works/pi-ai/compat": _bundledPiAiCompat, // line 42
```

The Bun-binary path is a plain object lookup — subpath keys work trivially. So the "jiti ignores subpath aliases containing `/`" theory would only affect the Node/jiti path, and there the loader explicitly passes the alias to `createJiti(..., { alias: getAliases() })`. In my install that alias resolves the subpath.

## It's also a no-op even if the bug were real

Per the loader's own comments, the bare `@earendil-works/pi-ai` is aliased to the compat entrypoint as a "strict superset of the core entrypoint." So bare-name and `/compat` resolve to **the same module**. If `/compat` resolution were genuinely broken, swapping the specifier wouldn't fix it — the loader (or jiti) would.

## What I'd want before merging

Could you share:
1. `pi-coding-agent --version` (is the host older than the alias registration?)
2. Whether it reproduces after clearing the jiti cache / a fresh `pi install`

Most plausible explanations for the symptom, none of which the PR addresses:
- **Stale jiti cache** — cached transpiled modules can produce exactly this path-append artifact; a cache clear often fixes it.
- **Old host** — pre-0.80 hosts may not register the `/compat` alias key.
- **Broken `@earendil-works/pi-ai` install** — missing the `./compat` exports map.

Happy to be wrong here — if there's a real jiti 2.7.0 subpath-alias bug, a minimal repro (just the alias + a one-line import, no pi needed) would settle it quickly. But on the current evidence I don't think this change fixes a real bug, and it moves away from the explicit, self-documenting `/compat` specifier.

chronoAP:
I'lll have a look on my end when I get a second - could just be a local issue on my machine.

nicobailon:
Thanks for chasing this down. Current main intentionally keeps the `@earendil-works/pi-ai/compat` import path for the supported Pi 0.80.x runtime, and the reported 0.79.x failure appears obsolete after upgrading. This PR is also stale/conflicting now, so I’m closing it rather than merging the bare-package import change.

PR #120, opened 2026-07-07
fix: move sendCuratorFallbackUpdate into catch block to fix ReferenceError

### LINKED ISSUE #103: openCuratorBrowser catch block crashes pi with ReferenceError: sendCuratorFallbackUpdate is not defined when browser fails to launch
## `uncaughtException` in `openCuratorBrowser` catch block when browser fails to launch — kills the pi process

   ### Summary

   When the browser cannot be launched (e.g. no X11 display on headless Linux, `xdg-open` exit code 1, or `openInBrowser` throwing for any other reason), `openCuratorBrowser` throws a *second* unhandled error
 inside its own `catch` block and tears down the entire pi process.

   This is a different bug from #55, which was closed for the first-layer error. Closing #55 did not surface this second-layer crash; the only signal to the user is:

 ```

 Failed to open curator UI: Failed to open browser (exit code 1)
     at openCuratorBrowser (index.ts:1184:9)

 ```

   `web_search` is supposed to degrade gracefully when the curator UI can't be opened (it should at least print a fallback message and let the user open the URL manually). Instead, the agent process is killed.

   ### Environment

   - pi-web-access: 0.12.0 (also reproduced on `main` @ 0.13.0 — same code)
   - Node: any
   - OS: Linux, headless (no `$DISPLAY`, no `chromium`/`firefox` installed). Most likely reproduces on any system where `openInBrowser` throws.
   - pi-coding-agent: 0.37.3+

   ### Steps to reproduce

   1. On a headless Linux box (or any env where `xdg-open` cannot launch a browser), run pi with `pi-web-access` installed.
   2. Trigger a `web_search` call with the default `summary-review` workflow.
   3. The curator server starts, then `openInBrowser` throws because no browser is available.
   4. Control flows into the `catch` block at `index.ts:1183`.
   5. The `catch` block calls `sendCuratorFallbackUpdate(...)`.
   6. **The pi process exits with `uncaughtException: ReferenceError: sendCuratorFallbackUpdate is not defined`.**

   ### Root cause

   `sendCuratorFallbackUpdate` is declared as a `const` arrow function **inside the `try` block** at `index.ts:1135`, but it is invoked from the surrounding `catch` block at `index.ts:1184`.

   ```typescript
   // index.ts
   async function openCuratorBrowser(callId, pc, searchesComplete = true) {
       let handle = null;
       try {
           // ...
           const sendCuratorFallbackUpdate = (message) => {   // ← line 1135, block-scoped to `try`
               pc.onUpdate?.({ /* ...uses `handle.url`... */ });
           };
           // ...
           await openInBrowser(pi, handle.url);                // throws on headless
       } catch (err) {
           // ...
           sendCuratorFallbackUpdate("Search curator is running, but the browser did not open automatically.");  // ← line 1184: ReferenceError
       }
   }
 ```

 const is block-scoped to the try block. When the exception jumps to catch, the binding is out of scope, so the lookup falls through to the global scope and throws ReferenceError. That ReferenceError is
 uncaught by anyone (the catch only handles the original openInBrowser error), bubbles to Node's process-level uncaughtException, and pi exits.

 The if (handle && ...) guard on line 1183 already ensures the fallback is only called when handle is non-null, so the binding is safe to move.

 ### Suggested fix

 Move the const sendCuratorFallbackUpdate declaration out of the try block, to function scope (or convert it to a function declaration which hoists):

 ```diff
    async function openCuratorBrowser(callId: string, pc: PendingCurate, searchesComplete = true): Promise<void> {
        if (pendingCurates.get(callId) !== pc) return;
        let handle: CuratorServerHandle | null = null;
   +    const sendCuratorFallbackUpdate = (message: string) => {
   +        pc.onUpdate?.({
   +            content: [{ type: "text", text: `${message}\nOpen manually: ${handle?.url ?? "(unavailable)"}` }],
   +            details: {
   +                phase: "curator-fallback",
   +                progress: searchesComplete ? 1 : 0.5,
   +                curatorUrl: handle?.url,
   +                timeoutSeconds: pc.timeoutSeconds,
   +                shortcut: curateKey,
   +                browserOpenError: pc.browserOpenError,
   +            },
   +        });
   +    };
        try {
            pc.phase = "curating";
            // ...
   -        const sendCuratorFallbackUpdate = (message: string) => { ... };
            // ...
            await openInBrowser(pi, handle.url);
        } catch (err) { ... }
    }
 ```

 Two notes for the reviewer:
 1. handle is captured by closure; the if (handle && ...) guard at the call site already ensures it is non-null when this function runs, but the handle?.url and handle?.url ?? "(unavailable)" above also defend
    against the (currently impossible) case where startCuratorServer itself throws and we still want a graceful message.
 2. A function declaration would also work and would naturally hoist.

 ### Workaround (no code change)

 Until fixed, users on headless / WSL / SSH-without-X-forwarding machines can avoid the crash by setting:

 ```json
   // ~/.pi/web-search.json
   {
     "workflow": "auto-summary"
   }
 ```

 This bypasses openCuratorBrowser entirely (only summary-review opens the browser; auto-summary returns the AI summary directly). The buggy code path is never reached, but the curator UI is no longer available
 in this mode.

 ### Why this should be high priority

 This bug turns a recoverable UX problem (the curator page couldn't open) into a process-level crash. A single failed web_search call kills the agent mid-conversation, and the user is dropped back to the shell
 with no warning other than the stack trace. For users on headless servers (a deployment case that #18, #19, #20, and #55 all explicitly call out) the extension is effectively unusable in its default
 configuration.

2seoik:
## Summary

Fixes a `ReferenceError: sendCuratorFallbackUpdate is not defined` crash that occurs when the curator browser fails to open on non-macOS platforms.

## Root Cause

The `sendCuratorFallbackUpdate` helper was defined with `const` inside the `try` block of `openCuratorBrowser()`, but it is called exclusively from the `catch` block. Since `const`/`let` are block-scoped in JavaScript, the function is not accessible from `catch`, causing an unhandled `ReferenceError`.

`sendCuratorFallbackUpdate` is never invoked inside the `try` block — it was misplaced there, likely a leftover from a refactoring iteration.

## Changes

1. **Moved** the `sendCuratorFallbackUpdate` definition from the `try` block into the `catch` block where it is actually called.
2. **Changed `handle.url` → `handle?.url`** with optional chaining as a defensive safeguard. The current call site is already gated behind a `handle && ...` check, so this isn't required to fix the crash itself, but it guards against future call sites where `handle` could still be `null` (e.g. if the catch is entered before the curator server finishes starting).

I confirmed via inspection that `sendCuratorFallbackUpdate` has no other call sites in `openCuratorBrowser()` — it's referenced exactly once, in the `catch` block.

## Affected Platforms

This is not WSL-specific. Any environment where `openInBrowser()`'s platform branch fails — `xdg-open` missing/misconfigured on Linux, `open`/`cmd start` failing on macOS/Windows, headless/container environments with no browser at all — will hit this same `catch` path and crash. WSL2 is simply the environment I reproduced it in; macOS users are less likely to hit it in practice because the Glimpse window path returns early (`return`), skipping `openInBrowser` entirely, but a Glimpse open failure that falls through to `openInBrowser` would trigger it there too.

### Environment (where I reproduced it)

- OS: Windows 11 WSL2 (Ubuntu)
- Node.js: v24.16.0
- pi-web-access: v0.13.0

### Steps to Reproduce

1. Use pi agent in an environment without a working `xdg-open`/`open`/`cmd start` (e.g. WSL2 without a browser mapping, or headless Linux)
2. Run a web search (`web_search`) with curator browser (`workflow: summary-review`)
3. `openInBrowser()` throws → `catch` block runs → `ReferenceError` on `sendCuratorFallbackUpdate`

### Error

```
    at openCuratorBrowser (/pi-web-access/index.ts:1188:9)
```

### After Fix

The curator browser open failure is now handled gracefully — the fallback message is properly delivered to the user via `pc.onUpdate` instead of crashing.

### Test Coverage

No existing test covers the `openCuratorBrowser` catch/fallback path. I did not add one in this PR since I don't see a test harness for this module yet — happy to add a regression test (e.g. mocking `openInBrowser` to reject and asserting `onUpdate` is called instead of throwing) if the maintainer points me to the preferred testing setup/conventions for this repo.

Closes #103

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -1136,20 +1136,6 @@ export default function (pi: ExtensionAPI) {
 			}
 			if (searchesComplete) handle.searchesDone();
 
-			const sendCuratorFallbackUpdate = (message: string) => {
-				pc.onUpdate?.({
-					content: [{ type: "text", text: `${message}\nOpen manually: ${handle.url}` }],
-					details: {
-						phase: "curator-fallback",
-						progress: searchesComplete ? 1 : 0.5,
-						curatorUrl: handle.url,
-						timeoutSeconds: pc.timeoutSeconds,
-						shortcut: curateKey,
-						browserOpenError: pc.browserOpenError,
-					},
-				});
-			};
-
 			pc.onUpdate?.({
 				content: [{ type: "text", text: searchesComplete ? "Waiting for summary approval in browser..." : "Searches streaming to browser..." }],
 				details: {
@@ -1181,6 +1167,19 @@ export default function (pi: ExtensionAPI) {
 			}
 			await openInBrowser(pi, handle.url);
 		} catch (err) {
+			const sendCuratorFallbackUpdate = (message: string) => {
+				pc.onUpdate?.({
+					content: [{ type: "text", text: `${message}\nOpen manually: ${handle?.url}` }],
+					details: {
+						phase: "curator-fallback",
+						progress: searchesComplete ? 1 : 0.5,
+						curatorUrl: handle?.url,
+						timeoutSeconds: pc.timeoutSeconds,
+						shortcut: curateKey,
+						browserOpenError: pc.browserOpenError,
+					},
+				});
+			};
 			const message = err instanceof Error ? err.message : String(err);
 			console.error(`Failed to open curator UI: ${message}`);
 			if (handle && activeCurators.get(callId) === handle && pendingCurates.get(callId) === pc) {


nicobailon:
Thanks for the fix. The curator fallback scope bug is now fixed on main via #147, using a function-scope helper plus null guard, so I’m closing this as superseded. Credit for this related PR is included in the changelog.

PR #121, opened 2026-07-07
fix: hoist curator fallback helper so the catch block can reach it

pnisarg:
## Summary

Fixes a `ReferenceError` crash that takes down the entire Pi process when the search curator's browser fails to auto-open — the common case in Docker, WSL, SSH, and other headless environments.

## Root cause

In `openCuratorBrowser` (`index.ts`), the `sendCuratorFallbackUpdate` helper is declared with `const` **inside** the `try` block, but it is invoked from the sibling `catch` block:

```ts
try {
  ...
  const sendCuratorFallbackUpdate = (message) => { ... }; // declared in try scope
  ...
  await openInBrowser(pi, handle.url);                    // throws in headless envs
} catch (err) {
  ...
  sendCuratorFallbackUpdate("...browser did not open..."); // not in scope here
}
```

`const`/`let` are block-scoped, so the identifier does not exist in the `catch` block. When `openInBrowser` throws (no display/browser available), the `catch` handler runs the intended graceful-fallback path and instead throws:

```
    at openCuratorBrowser (.../pi-web-access/index.ts)
```

Because this executes on an async tick with no surrounding handler, Pi treats it as an `uncaughtException` and exits. The error is masked on desktops where the browser opens successfully and the `catch` block never runs — it only manifests in headless environments, which is exactly where the fallback was meant to help.

Regression introduced in b50d118 (`fix: keep curator alive when browser open fails`).

## Fix

- Hoist `sendCuratorFallbackUpdate` to function scope (before the `try`) so both the `try` and `catch` blocks can reference it.
- Add an `if (!handle) return;` guard, since the helper is now declared before `handle` is assigned. The only caller already guards with `if (handle && ...)`, so behavior is unchanged.

## Tests

- Added a regression test in `test/curator-fallback.test.mjs` asserting the helper is declared before the `try` block. Verified it **fails** against current `main` and **passes** with this fix.
- Full suite green locally with dependencies installed (`npm test` → 70/70 pass).

## Reproduction

Run `web_search` (with the default browser-curator workflow) on any host with no reachable browser (Docker/WSL/SSH/headless). Before this fix, Pi crashes with the `ReferenceError` above instead of printing the manual curator URL.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Fixed
+- Fixed a `ReferenceError: sendCuratorFallbackUpdate is not defined` crash that took down the whole process when the curator browser failed to auto-open (Docker, WSL, SSH, and headless environments). The fallback helper was declared inside the `try` block but called from the sibling `catch` block; it is now declared at function scope so the browser-open failure path resolves correctly.
+
 ## [0.13.0] - 2026-06-25
 
 ### Added
--- a/index.ts
+++ b/index.ts
@@ -964,6 +964,20 @@ export default function (pi: ExtensionAPI) {
 	async function openCuratorBrowser(callId: string, pc: PendingCurate, searchesComplete = true): Promise<void> {
 		if (pendingCurates.get(callId) !== pc) return;
 		let handle: CuratorServerHandle | null = null;
+		const sendCuratorFallbackUpdate = (message: string) => {
+			if (!handle) return;
+			pc.onUpdate?.({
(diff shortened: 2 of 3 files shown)


BeGild:
这个修复方案很好！加上了 handle 的空值守卫，并且有完整的回归测试和 CHANGELOG。

我在无头环境中也遇到了同样的 ReferenceError 崩溃，这个修复确实能解决问题。希望尽快合并！ 👍

nicobailon:
Thanks for the fix. I landed the same helper-scope repair in #147, with the regression test and changelog credit preserved, so I’m closing this PR as superseded by the maintainer replacement.

PR #123, opened 2026-07-08
feat: add Firecrawl as self-hosted search and content extraction provider

fank:
## Summary

Adds **Firecrawl** as an optional self-hosted provider for `web_search` and `fetch_content`. Addresses the Firecrawl portion of #105.

Rebased onto current `main` (`69f9772`, including the SearXNG work from #107) as a single commit. No breaking changes: Firecrawl is inert unless `firecrawlBaseUrl` / `FIRECRAWL_BASE_URL` is set.

## Review points from the previous round

**`/v1` vs `/v2` — v2 is the default.** Firecrawl's current API is v2 and self-hosted instances expose it, which matches what #105 asked for. `firecrawlApiVersion` / `FIRECRAWL_API_VERSION` accepts `"v1"` for older self-hosted images; anything else throws instead of silently hitting a wrong endpoint. Both search response shapes are handled — v1 returns a flat `data` array, v2 returns `data.web` / `data.news`. Domain filters are sent as v2 `includeDomains`/`excludeDomains` *and* re-applied client-side, since v1 has no such parameters and self-hosted backends do not always honour the v2 ones.

**Errors are preserved, not swallowed.** `firecrawlFetch()` throws on transport failure, non-2xx, invalid JSON, an unexpected envelope shape, and `success: false`. `extractWithFirecrawl()` returns `null` only when the scrape produced no usable Markdown. `extractContent()` catches the throw and surfaces it as `Firecrawl fallback failed: …` in its guidance, the same way the Parallel fallback already does. Config-parse errors short-circuit, matching the existing providers.

**SSRF — same contract as SearXNG.** Requests go through `fetchRemoteUrl` with `loadSsrfConfig()`, so the Firecrawl instance and its redirects are guarded exactly like SearXNG's after #107: a self-hosted instance on a private address needs its range in `ssrf.allowRanges`, and `localhost` is never reachable (use `127.0.0.1`). *Separately*, the scrape target is validated with `validateRemoteUrl` inside the provider as well as in `extractContent` — Firecrawl fetches server-side, so handing it a private URL would launder the target past the local guard even when the instance is legitimately private.

> Note for anyone following along: an earlier revision of this branch exempted the Firecrawl base URL from the guard. That predates #107 landing; it now follows the SearXNG contract instead.

**Search in auto mode.** Kept, but moved to *after* Exa rather than before it. A base URL alone does not mean search works: self-hosted `/search` needs its own search backend configured in the Firecrawl deployment, so someone who set `firecrawlBaseUrl` purely for extraction (the literal #105 use case) still gets Exa first. Explicit `provider: "firecrawl"` always goes direct. Happy to drop the search half entirely if you'd rather keep this PR to extraction only.

**Recency.** Mapped to Firecrawl `tbs` (`day`→`qdr:d` … `year`→`qdr:y`). A value that cannot be expressed is dropped rather than sent, and this is asserted in a test.

## Auth

Bearer token via `firecrawlApiKey` / `FIRECRAWL_API_KEY`; omitted entirely when unset, since self-hosted instances commonly run without auth. The `FIRECRAWL_BASIC_AUTH` handling advertised in the old description was removed before the previous review and is not in this branch.

## Tests

14 new tests in `test/firecrawl.test.mjs`:

- default v2 search endpoint, Bearer header, `tbs`, `sources`, `includeDomains`/`excludeDomains`, and v2 `data.web` mapping
- v1 search: `/v1/search`, flat `data` array, client-side-only domain filtering
- v1 scrape: `/v1/scrape` and v1 envelope parsing
- unsupported `firecrawlApiVersion` throws and issues no request
- config-file base URL and key, trailing-slash normalization, no `Authorization` header when unkeyed
- unconfigured: `isFirecrawlAvailable()` is `false`, search and extract both throw, no request made
- unsupported recency value sends no `tbs`
- extraction through a loopback instance works once its range is in `ssrf.allowRanges`, and is refused when it is not
- private targets (`127.0.0.1`, `localhost`, `169.254.169.254`, and a public hostname resolving into `10.0.0.0/8`) throw with **zero** fetch calls
- `fetch_content` on a private URL never reaches Firecrawl
- response-shape failures: 502, HTML instead of JSON, `success: false`, non-object envelope — each surfaces a distinct error
- empty Markdown returns `null`
- malformed `ssrf.allowRanges` fails loud rather than silently skipping the provider

Suite: 159/160 pass. The one failure (`failed password lookups are retried instead of cached`, Chromium cookie/keyring) also fails on unmodified `main` in this environment and is unrelated.

Also scrubs `FIRECRAWL_*` from the existing provider test harnesses — with a real Firecrawl configured in the developer's environment it leaked into `search-providers` assertions.

Refs #105


WAY29:
Any review? want this feature

nicobailon:
Thanks for the Firecrawl work. I reviewed head `e6a5cf8b0a6236e23dcecdd32282a4367527b84f` against current main and issue #105. I’m not going to merge this head as-is, but it’s useful source material.

The main blockers are scope and API contract. #105 asked mostly for Firecrawl extraction and mentions `/v2/scrape`; this PR adds search plus `/v1/search` and `/v1/scrape`. We need to decide which Firecrawl versions/surfaces are supported, whether Firecrawl search belongs in auto mode, and how extraction errors should be preserved instead of being converted to silent fallback. The direct provider functions also need current-main SSRF expectations reviewed and tested.

If you revise, please rebase onto current main, clarify `/v1` vs `/v2`, add tests for config/auth, unsupported recency behavior, private-target non-invocation through `fetch_content`, and response-shape failures. Credit for your PR will be preserved if this lands or is clean-ported.

fank:
Revised and force-pushed — clean re-port onto current `main` (`61dbfc6`), one commit. Full details in the updated description; the short version against your points:

- **Rebase**: done, single commit rather than the old self-reverting three.
- **`/v1` vs `/v2`**: `v2` is now the default (matches #105 and what self-hosted instances expose). `firecrawlApiVersion` opts back to `v1` for older images; anything else throws. Both search response shapes (`data[]` on v1, `data.web`/`data.news` on v2) and both scrape endpoints are covered by tests.
- **Extraction errors**: no longer converted to silent fallback. HTTP status, invalid JSON, unexpected envelope shape and `success: false` all throw and land in `extractContent`'s guidance as `Firecrawl fallback failed: …`, mirroring the Parallel path. `null` now means only "scrape returned no Markdown".
- **SSRF**: the scrape target is validated inside the provider as well as in `extractContent`, with `ssrf.allowRanges`/`trustEnvProxy` threaded through and strict defaults for direct calls. Tests assert **zero** fetch calls for loopback, `localhost`, link-local, and a public hostname resolving into `10.0.0.0/8`, plus a `fetch_content` end-to-end case. The Firecrawl base URL stays exempt so private self-hosted instances still work — also tested.
- **Search in auto mode**: kept, but moved to *after* Exa. A base URL alone doesn't imply working search — self-hosted `/search` needs its own backend configured — so someone who set `firecrawlBaseUrl` only for extraction still gets Exa first. Say the word and I'll cut the search half entirely and leave this as extraction-only.
- **Config/auth + recency**: covered. Bearer from env or config file, no `Authorization` header when unkeyed, trailing-slash normalization, unconfigured throws without a request. Recency maps to `tbs`; an inexpressible value is dropped rather than sent.

12 new tests; suite is 145/146 — the single failure (`failed password lookups are retried instead of cached`) also fails on unmodified `main` here and is unrelated. One drive-by: the existing provider test harnesses didn't scrub `FIRECRAWL_*`, so a real local Firecrawl config leaked into unrelated assertions; that's fixed.

fank:
Rebased onto current `main` (`69f9772`) after the SearXNG merge — still one commit, force-pushed.

One deliberate behavior change while resolving: **Firecrawl now follows the same SSRF contract as SearXNG.** Requests go through `fetchRemoteUrl` with `loadSsrfConfig()`, so the instance's own base URL and redirects are guarded, and a private instance needs its range in `ssrf.allowRanges` (`localhost` is never reachable — use `127.0.0.1`). This supersedes what I wrote in my previous comment about the base URL being exempt; that predates #107 landing, and matching your SearXNG contract is clearly the right call. The scrape *target* is still validated separately inside the provider, since Firecrawl fetches server-side.

Also merged cleanly alongside SearXNG in the provider surfaces: `SearchProvider`, curator buttons/labels, `availableProviders`, `source_check`'s provider enum, and the "no search provider available" guidance now list both. Firecrawl sits after Exa in the auto chain rather than up front with SearXNG, for the reason in the code comment: SearXNG is a search engine, but a Firecrawl base URL is usually set for extraction and its `/search` only answers when the deployment has its own backend.

14 Firecrawl tests, suite 159/160. The single failure (`failed password lookups are retried instead of cached`) also fails on unmodified `main` here — Chromium keyring, unrelated.

nicobailon:
Thanks again for the Firecrawl work here, Florian. I split the provider backlog and landed the extraction half in #165 with credit to you in the changelog and PR body.

I intentionally left Firecrawl search out for now because the replacement keeps the initial Firecrawl surface to content extraction only, with cache-only-by-default scraping and an explicit fresh-scrape trust boundary. Closing this PR as superseded by #165.

PR #125, opened 2026-07-10
Improve startup time: Defer content extraction until first use

kaushikgopal:
## Change Brief

At the most basic level: this PR is a startup performance change, not a change to extraction behavior or tool availability.

## Summary

- Replaces the eager `extract.ts` runtime import with a memoized dynamic import.
- Keeps `web_search`, `fetch_content`, `get_search_content`, commands, shortcuts, and session handlers registered eagerly.
- Defers the Readability, LinkeDOM, Turndown, PDF, and video extraction graph until `fetch_content` or a background content fetch first needs it.
- Adds a regression test that observes module loading directly and exercises first-use HTML extraction against a local server.

## Why

Importing the extension eagerly loaded the full content-extraction dependency graph even when a session only needed search. Deferring that graph removes most of the extension's startup overhead while preserving the existing fetch path and sharing one import promise across concurrent first-use calls.

## Before / After

Before, loading the extension also loaded content extraction and its heavy transitive dependencies before tool registration completed.

After, registrations remain synchronous and immediately available. The extraction graph loads once, on the first direct or background content fetch, and then remains cached.

Measured isolated startup overhead:

| Environment | Before | After |
| --- | ---: | ---: |
| Current | 66 ms | 15 ms |
| Home | 77 ms | 16 ms |
| Direct import | ~65 ms | ~12 ms |

## Specific Changes

- **Startup path**: use type-only extraction imports and a memoized `import("./extract.ts")` wrapper.
- **Regression coverage**: verify eager registration without loading `extract.ts`, then invoke `fetch_content` and verify both module loading and extracted article content without timing assertions.

## Commit Notes

- `perf: lazy-load content extraction`
  - Why: avoid paying extraction-only initialization costs during every extension startup.
  - What changed: defer and memoize the extraction module import, with first-use integration coverage.
  - Implication: the first content fetch pays the module-load cost; later fetches reuse the loaded module.

## Testing

- `npm test` — 70 tests passed.
- Typecheck — no typecheck script or `tsconfig.json` is configured in the package.

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -2,7 +2,7 @@ import type { ExtensionAPI, ExtensionContext } from "@earendil-works/pi-coding-a
 import { Box, Text, truncateToWidth } from "@earendil-works/pi-tui";
 import { Type } from "typebox";
 import { StringEnum, complete, type Model } from "@earendil-works/pi-ai/compat";
-import { fetchAllContent, type ExtractedContent } from "./extract.ts";
+import type { ExtractedContent, ExtractOptions } from "./extract.ts";
 import { normalizeFetchContentParams } from "./fetch-params.ts";
 import { clearCloneCache } from "./github-extract.ts";
 import { search, type SearchProvider, type ResolvedSearchProvider } from "./gemini-search.ts";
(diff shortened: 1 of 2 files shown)


nicobailon:
Thanks for this startup improvement. I landed it in #142 with changelog credit, keeping your authored commit in the maintainer branch.

PR #129, opened 2026-07-14
feat: Support searchModel config for openai search

gshahbazian:
So users can use gpt-5.6 to run the search


nicobailon:
Thanks for the OpenAI search-model proposal. I reviewed head `86c79aaf5238486a3b2977e942308b7554512d7a` against current main. It is not covered by current source, but I’m not merging this head as-is because the model configuration contract needs a product/API decision and it overlaps with #131.

Today `searchModel` is documented and implemented as a Gemini search override. Extending it to OpenAI/Codex raises a few decisions: whether bare Gemini IDs and provider-qualified OpenAI IDs should share one field, whether unavailable pins should hard-fail or fall back, and whether arbitrary raw `openai/<id>` values should be accepted differently from registry-backed `openai-codex/<id>` values.

If we proceed, I’d prefer an explicit provider-qualified contract with focused tests for API-key and Codex registry paths, invalid/unavailable pins, request body model selection, and auto-provider fallback behavior. We’ll preserve credit for this proposal if that design lands.

PR #131, opened 2026-07-15
Add configurable OpenAI and Perplexity gateway settings

robobryce:
## Summary
- support `OPENAI_BASE_URL` / `openaiBaseUrl` for OpenAI Responses search
- support `OPENAI_SEARCH_MODEL` / `openaiSearchModel`
- support `PERPLEXITY_BASE_URL` / `perplexityBaseUrl`
- support `PERPLEXITY_MODEL` / `perplexityModel`

## Testing
- `node --test test/gateway-base-url.test.mjs`


nicobailon:
Thanks for the gateway-config work. I reviewed head `35f229561375d9223e0bc26fa6fa9c0df924d9c0` against current main. I’m not going to merge it as-is because the OpenAI contract is ambiguous after the request-time credential/model-registry changes.

The Perplexity base URL/model portion is a clean additive idea. The OpenAI part only reliably applies to the literal API-key fallback path; normal context-resolved OpenAI/Codex auth can still choose the fixed Codex endpoint/model, which conflicts with the broad README wording. This also overlaps with #129’s `searchModel` proposal, so we should not land both model-setting designs independently.

A good next revision would define one canonical OpenAI search-model/base-url contract, cover context-auth/Codex preservation in tests, keep the Perplexity changes, add changelog credit, and document that #154-style credential sources are not automatically supported for these new fields unless intentionally added.

nicobailon:
Thanks for this config work. I reviewed it against current main after the credential-source and provider-routing changes. The Perplexity base URL/model portion still looks like a useful additive slice, but the OpenAI side needs a single canonical contract before it can land.

Current OpenAI search has both Pi/Codex registry auth and API-key auth. A normal OpenAI gateway/base URL cannot be documented as applying to Codex subscription auth unless that path is explicitly supported and tested. Please preserve the Perplexity idea for a smaller rework, and either drop the OpenAI fields for now or redesign them around provider-qualified model/base URL semantics with tests for API-key auth, Codex registry auth, unavailable model pins, request body selection, and auto fallback behavior.

nicobailon:
I did a fresh pass after the provider/security changes landed. The Perplexity model/base-URL slice still looks salvageable, but the current patch has one hard blocker: configurable OpenAI/Perplexity base URLs would send bearer credentials to arbitrary configured hosts without the SSRF/redirect validation we now require for configurable remote endpoints.

The OpenAI half also needs a sharper contract. In normal Pi/Codex context-auth usage, the provider selects registry auth/model and keeps using the Codex endpoint, so broad `OPENAI_BASE_URL` / `OPENAI_SEARCH_MODEL` wording is misleading unless the fields are explicitly scoped to API-key fallback or a new provider-qualified contract is designed.

A rework should define that contract first, then add SSRF/private-host and redirect tests, credential redaction tests, context-auth preservation tests, and clear Cloudflare-gateway wording rather than implying the Gemini `CLOUDFLARE_API_KEY` path applies here.

nicobailon:
Thanks again for the gateway configuration work, RoboBryce. I’m closing this PR because the current patch is not mergeable under the extension’s present credential and network-safety contracts: configurable credential-bearing endpoints need SSRF and redirect protection, and the OpenAI fields need an explicit API-key-versus-Codex context-auth contract.

The Perplexity configuration idea remains useful. If you address those blockers on current `main`, feel free to request that this PR be reopened or open a fresh focused PR, and I’ll reconsider it.

PR #132, opened 2026-07-17
Fix web activity shortcut widget registration

treyhoover:
## Summary

Pass a component factory to `ctx.ui.setWidget` for the Ctrl+Shift+W web activity widget instead of passing a `Text` instance directly.

Pi's current widget API accepts either a string array or a component factory. Passing the component instance reaches the factory branch and throws `content is not a function`.

## Validation

- `node --test test/tool-registration-config.test.mjs`
- `npm test` — 70 tests passed


nicobailon:
Thanks for this fix. I landed it in #142 with changelog credit, keeping your authored commit in the maintainer branch.

PR #133, opened 2026-07-17
fix: prevent pi crash when curator browser fails to open

### LINKED ISSUE #103: openCuratorBrowser catch block crashes pi with ReferenceError: sendCuratorFallbackUpdate is not defined when browser fails to launch

### LINKED ISSUE #116: ReferenceError: sendCuratorFallbackUpdate is not defined in openCuratorBrowser
## Environment
   - pi-web-access version: 0.13.0
   - OS: Linux (running as root)
   - Triggered when: browser fails to open the curator UI (e.g. Chromium refuses to start without --no-sandbox under root)

   ## Description
   `openCuratorBrowser` crashes with `ReferenceError: sendCuratorFallbackUpdate is not defined` when the browser fails to open.

   ## Root cause
   In `index.ts`, `sendCuratorFallbackUpdate` is declared with `const` inside the `try` block (around line 1139), but it is referenced inside the corresponding `catch` block (around line
 1188). `const` is block-scoped, so the `catch` block cannot see it.

   ```ts
   try {
       // ...
       const sendCuratorFallbackUpdate = (message: string) => { /* ... */ };
       // ...
       await openInBrowser(pi, handle.url);
   } catch (err) {
       // ...
       sendCuratorFallbackUpdate("Search curator is running, but the browser did not open automatically.");
       // ^^^ ReferenceError
   }
 ```

 Suggested fix

 Move sendCuratorFallbackUpdate definition above the try block (it only depends on pc, handle, searchesComplete, curateKey, which are all available in the outer scope), or inline the
 pc.onUpdate?.(...) call inside the catch block.

 Impact

 When the curator browser fails to open, pi crashes with an uncaught exception instead of gracefully falling back to a manual-open message.

### LINKED ISSUE #126: Crashses pi with sendCuratorFallbackUpdate is not defined
Failed to open curator UI: /usr/bin/xdg-open: 1032: www-browser: not found
/usr/bin/xdg-open: 1032: links2: not found────────────────────────────────────────────────────────────────────────────────────────────────
/usr/bin/xdg-open: 1032: elinks: not found
/usr/bin/xdg-open: 1032: links: not found                                                               (ollama) qwen3.6-27b-262144:latest
/usr/bin/xdg-open: 1032: lynx: not found
/usr/bin/xdg-open: 1032: w3m: not found
xdg-open: no method available for opening 'http://localhost:37603/?session=c36037bc-3c69-4b26-82dd-137173a201c3'

    at openCuratorBrowser (/home/tonydiep/.pi/agent/npm/node_modules/pi-web-access/index.ts:1188:9)
    at process.processTicksAndRejections (node:internal/process/task_queues:104:5)
t

### LINKED ISSUE #127: web_search crashes pi with `ReferenceError: sendCuratorFallbackUpdate is not defined` in headless / no-browser environments
## Summary
On a headless host (no GUI/text browser installed), the default `web_search` workflow (`summary-review`) tries to open the interactive curator UI via `xdg-open`, which fails. The failure handler then throws `ReferenceError: sendCuratorFallbackUpdate is not defined`, which propagates as an uncaught exception and **terminates the entire pi process**.

## Environment
- `pi-web-access`: 0.13.0
- pi running over the Telegram bridge (no TUI / headless Linux)
- No browser available — `xdg-open` reports `www-browser`, `links2`, `elinks`, `links`, `lynx`, `w3m` all missing

## Error
```
Failed to open curator UI: /usr/bin/xdg-open: 882: www-browser: not found
... (links2, elinks, links, lynx, w3m: not found)
xdg-open: no method available for opening 'http://localhost:PORT/...'

    at openCuratorBrowser (/.../pi-web-access/index.ts:1188:9)
    at process.processTicksAndRejections (node:internal/process/task_queues:103:5)
```

## Root cause
Two distinct issues combine to turn a recoverable "browser didn't open" situation into a process crash.

### 1. Scope bug in `openCuratorBrowser`
`sendCuratorFallbackUpdate` is declared with `const` **inside the `try` block** (~line 1139):
```ts
			const sendCuratorFallbackUpdate = (message: string) => {
				pc.onUpdate?.({ ... });
			};
```
but it is **referenced inside the sibling `catch` block** (line 1188):
```ts
		} catch (err) {
			const message = err instanceof Error ? err.message : String(err);
			console.error(`Failed to open curator UI: ${message}`);
			if (handle && activeCurators.get(callId) === handle && pendingCurates.get(callId) === pc) {
				pc.browserOpenError = message;
				sendCuratorFallbackUpdate("Search curator is running, but the browser did not open automatically."); // ← not in scope
			}
```
A `const` declared in a `try` block is not visible in the sibling `catch` block, so the fallback intended to recover from a failed `openInBrowser` instead throws `ReferenceError`.

### 2. `hasUI` detection treats `undefined` as "has UI"
In `resolveWorkflow` and its call site:
```ts
function resolveWorkflow(input: unknown, hasUI: boolean): WebSearchWorkflow {
	const normalized = typeof input === "string" ? input.trim().toLowerCase() : "";
	if (normalized === "auto-summary") return "auto-summary";
	if (!hasUI) return "none";
	if (normalized === "none") return "none";
	return "summary-review";
}
// call site (line 1273):
const workflow = resolveWorkflow(params.workflow ?? configWorkflow, ctx?.hasUI !== false);
```
`ctx?.hasUI !== false` evaluates to `true` when `ctx.hasUI` is `undefined`, so headless/bridge contexts (where `hasUI` isn't explicitly set to `false`) resolve to `"summary-review"` and attempt to open the curator browser even though no UI is present.

## Reproduction
1. On a host with no GUI/text browser installed, run pi over the Telegram bridge (or any context where `ctx.hasUI` is not explicitly `false`).
2. Invoke the `web_search` tool without an explicit `workflow`.
3. pi attempts to open the curator UI, `xdg-open` fails, and pi exits with the `ReferenceError` above.

## Suggested fix
- **Scope bug**: hoist `sendCuratorFallbackUpdate` to the top of `openCuratorBrowser` (or make the `catch` handler not depend on try-scoped state), so a failed `openInBrowser` is handled gracefully instead of crashing.
- **hasUI detection**: change `ctx?.hasUI !== false` → `ctx?.hasUI === true`, so `undefined` defaults to "no UI" (defensive default for headless/bridge environments).

## Workaround
Set the default workflow to `"none"` (or `"auto-summary"`) in `~/.pi/web-search.json`:
```json
{ "workflow": "none" }
```
This avoids the curator browser path entirely.

Happy to open a PR if this analysis looks right.

apoapostolov:
## Summary

When `openInBrowser` / `xdg-open` fails (headless Linux, missing browser, no display, etc.), `openCuratorBrowser`'s recovery path currently throws a second error and **kills the entire pi process**:

```
    at openCuratorBrowser (index.ts:1188:9)
```

`sendCuratorFallbackUpdate` was declared with `const` **inside the `try` block**, but invoked from the sibling **`catch`**. `const` is block-scoped, so the catch path cannot see it.

This PR hoists the helper to function scope so a failed browser open degrades to the intended fallback message (manual curator URL) instead of an uncaught exception.

Fixes #103  
Fixes #116  
Fixes #126  
Fixes #127  

## Change

```diff
 async function openCuratorBrowser(...) {
   let handle = null;
+  const sendCuratorFallbackUpdate = (message) => {
+    if (!handle) return;
+    pc.onUpdate?.({ /* manual URL fallback */ });
+  };
   try {
-    const sendCuratorFallbackUpdate = (message) => { ... };
     await openInBrowser(pi, handle.url);
   } catch (err) {
     sendCuratorFallbackUpdate("..."); // now in scope
   }
 }
```

## Test plan

- [ ] On a machine where `xdg-open` cannot open a browser (or mock `openInBrowser` to throw), run `web_search` with default `summary-review` workflow
- [ ] Confirm pi **does not exit** with `uncaughtException`
- [ ] Confirm fallback update / manual curator URL is surfaced
- [ ] Confirm normal path still works when a browser is available
- [ ] Optional: temporary `throw new Error("boom")` after `activeCurators.set` still hits catch without `ReferenceError`

## Workaround (until merge)

```json
// ~/.pi/agent/web-search.json  (or ~/.pi/web-search.json)
{ "workflow": "none" }
```

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -964,6 +964,21 @@ export default function (pi: ExtensionAPI) {
 	async function openCuratorBrowser(callId: string, pc: PendingCurate, searchesComplete = true): Promise<void> {
 		if (pendingCurates.get(callId) !== pc) return;
 		let handle: CuratorServerHandle | null = null;
+		// Hoisted out of try so the catch path can call it (const in try is not visible in catch).
+		const sendCuratorFallbackUpdate = (message: string) => {
+			if (!handle) return;
+			pc.onUpdate?.({
+				content: [{ type: "text", text: `${message}\nOpen manually: ${handle.url}` }],
+				details: {
+					phase: "curator-fallback",
+					progress: searchesComplete ? 1 : 0.5,
+					curatorUrl: handle.url,
+					timeoutSeconds: pc.timeoutSeconds,
+					shortcut: curateKey,
+					browserOpenError: pc.browserOpenError,
+				},
+			});
+		};
 		try {
 			pc.phase = "curating";
 
@@ -1135,20 +1150,6 @@ export default function (pi: ExtensionAPI) {
 				}
 			}
 			if (searchesComplete) handle.searchesDone();
-
-			const sendCuratorFallbackUpdate = (message: string) => {
-				pc.onUpdate?.({
-					content: [{ type: "text", text: `${message}\nOpen manually: ${handle.url}` }],
-					details: {
-						phase: "curator-fallback",
-						progress: searchesComplete ? 1 : 0.5,
-						curatorUrl: handle.url,
-						timeoutSeconds: pc.timeoutSeconds,
-						shortcut: curateKey,
-						browserOpenError: pc.browserOpenError,
-					},
-				});
-			};
 
 			pc.onUpdate?.({
 				content: [{ type: "text", text: searchesComplete ? "Waiting for summary approval in browser..." : "Searches streaming to browser..." }],


nicobailon:
Thanks for the fix. The curator fallback scope bug is now fixed on main via #147, with credit for this related PR in the changelog, so I’m closing this as superseded.

PR #139, opened 2026-07-22
Add SERPdive search provider

edendalexis:
## What

Adds [SERPdive](https://serpdive.com) as a search provider next to Tavily — `SERPDIVE_API_KEY` env var or `serpdiveApiKey` config, `provider: "serpdive"`, curator UI button, and a slot in the auto fallback chain right after Tavily.

SERPdive returns extracted, answer-ready page content per result instead of snippets. On a [public 1,000-question benchmark](https://github.com/edendalexis/serpdive-benchmark) it won 60.7% of decided quality duels against Tavily's default (basic) search while returning 20.2% fewer tokens (1,001 vs 1,255 per query on average) at the same speed; the advanced Tavily tier was not tested. Free tier, no card.

Full disclosure: I built SERPdive. If you'd rather not carry another provider, close away — no hard feelings.

## How

`serpdive.ts` mirrors `tavily.ts` structurally. Differences, forced by the API surface (`POST /v1/search` with only `query`, `model`, `answer`, `max_results`):

- `answer: true` maps to the tool's synthesized-answer field (included in the request price)
- `includeContent: true` requests the `moby` model (full readable page content) and maps results into `inlineContent`; default is `mako` (key sentences)
- `recencyFilter` becomes a query hint (the engine reads the query; there is no time-range parameter)
- `domainFilter` is applied client-side on returned result URLs
- `numResults` is capped at the API maximum of 10

## Testing

`npm test`: 72/72 passing, including three new tests (bearer auth + body shape + answer/inlineContent mapping, the 10-result cap, and auto-fallback reaching SERPdive after unavailable earlier providers). CHANGELOG entry added under Unreleased.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ All notable changes to this project will be documented in this file.
 ## [Unreleased]
 
 ### Added
+- Added SERPdive search provider with request-time credential sources, a free-tier `krill` default that never spends on install, opt-in `mako`/`moby` retrieval depth, locally applied domain filters, and recency documented as a query hint rather than a guaranteed freshness filter. Thanks Eden d'Alexis (@edendalexis) for #139.
 - Added opt-in ordered search routing through `searchRouting.providers` with explicit transient, quota, and network fallback kinds; named providers remain strict, single-provider config retains precedence, and exhausted routes preserve per-provider diagnostics. Thanks @smithyyang for #77.
 - Added optional self-hosted SearXNG search with SSRF-guarded base URLs, local-first auto selection, result filters, and documented `ssrf.allowRanges` opt-ins. Thanks to Marcos A. Núñez (@marnunez) for PR #107 and Avinash Kanaujiya (@avinashkanaujiya) for issue #105.
 - Added optional self-hosted Firecrawl extraction for `fetch_content`, using `/v2/scrape` by default (or `/v1/scrape` for older images), request-time credential sources, cache-only requests by default, cross-origin redirect credential stripping, and local-first fallback behavior when configured. Thanks Florian Kinder (@fank) for PR #123 and Avinash Kanaujiya (@avinashkanaujiya) for issue #105.
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 
 # Pi Web Access
 
-**Web search, content extraction, and video understanding for Pi agent. OpenAI/Codex search, zero-config Exa search, Brave, Parallel, Tavily, self-hosted SearXNG, optional browser-cookie Gemini Web, or bring your own API keys.**
+**Web search, content extraction, and video understanding for Pi agent. OpenAI/Codex search, zero-config Exa search, Brave, Parallel, Tavily, SERPdive, self-hosted SearXNG, optional browser-cookie Gemini Web, or bring your own API keys.**
 
 [![npm version](https://img.shields.io/npm/v/pi-web-access?style=for-the-badge)](https://www.npmjs.com/package/pi-web-access)
 [![License: MIT](https://img.shields.io/badge/License-MIT-yellow.svg?style=for-the-badge)](https://opensource.org/licenses/MIT)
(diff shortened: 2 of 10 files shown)


edendalexis:
Update while this is open: SERPdive now has a free tier, which changes the calculus for this integration.

`model: "krill"` is **free and unlimited under fair use** — no card, no credits to spend, no trial clock. It returns about 700 tokens per search (roughly half what the usual alternatives send), one request at a time, at low priority. The paid models are unchanged.

Why it matters here: the usual objection to adding a search provider is that reviewers and users need a paid key just to try it. That objection is gone — anyone can run this integration end to end for free, including whoever reviews this PR.

Nothing in this PR needs to change for that: the default stays `mako`, and `krill` is opt-in via the existing model parameter. Happy to add a line to the docs here if you'd like the free tier called out explicitly.

nicobailon:
Thanks for putting this together. I reviewed the current head `af18a65443bf3a311f1282fa6809058bcc3ad419` against current main. I’m not going to merge it as-is because it’s now behind a lot of provider/routing/credential changes and there are no completed checks on the exact head.

The integration shape looks plausible, but before this can move forward it needs a rebase onto current main, a full test run, and independent API-contract evidence for SERPdive’s response shape and model behavior (`mako`/`moby`, answer/results shape, max-results behavior, and error/abort cases). The recency behavior also needs to be documented as a query hint rather than a guaranteed freshness filter unless the API supports a real time bound.

If you rebase, please also reconcile the docs with the current provider list and add/refresh the focused tests on the current provider routing surface. We’ll keep credit for your PR if this lands or is reimplemented.

nicobailon:
Thanks for the SERPdive work. I did a current-main provider-policy pass and this is valuable enough to keep investigating rather than close by category: the answer-ready/inline-content angle and the reported fair-use/free model could be useful for people who choose it.

The branch is stale against current main, though, and before this can be considered it needs a rebase/current-main port that uses the shared request-time credential-source path, updates the current provider/schema/curator/docs surfaces, and preserves the existing routing/security behavior. Please also add independently reproducible contract/cost evidence for the selected model: request/response shape for default, mako/moby/krill if applicable, max results, HTTP errors, abort behavior, and whether recency is only a query hint. I do not want to silently ship a paid default if the intended value is the free/fair-use tier.

edendalexis:
Thanks for the detailed review — you were right on every count, including the one that mattered most. Rather than rebase `af18a65`, I ported the work onto current `main`: the branch was 52 commits behind and conflicted on every wiring point, so `serpdive.ts` is rewritten against today's `tavily.ts` shape instead of replayed.

**No paid default.** This was the right thing to block. The old revision sent `moby` for `includeContent` — SERPdive's *most* expensive model. The default is now `krill`, the free fair-use tier, so installing this provider never starts spending on a user's behalf. `mako` and `moby` are opt-in via `serpdiveModel` or `SERPDIVE_MODEL`, and an unrecognised value falls back to `krill` rather than throwing, so a typo in a config file cannot cost money either. `krill` has no answer synthesis, so `SearchResponse.answer` is assembled from the sources the same way `brave.ts` and `searxng.ts` already do for providers that do not synthesize.

**Credentials.** Now on the shared request-time path (`resolveCredential` / `hasCredentialSource`), so `$ENV_VAR` and `!command` sources work as they do everywhere else. The key is also redacted from every error path — the previous revision redacted none of them, which was a real leak on a 4xx body.

**Recency is a query hint.** Confirmed, and documented as such in the README and in the code: SERPdive exposes no time-range parameter, so `recencyFilter` is appended to the question and biases ranking. Results outside the window can still come back. Same story for `domainFilter`: there is no domain parameter, so the filter is applied locally to what came back — it can narrow a page of results, not ask the engine for more from a given domain. Both limits are asserted in the tests rather than left as prose.

**Contract evidence.** `evidence/CONTRACT-EVIDENCE.md` holds sixteen real calls against `api.serpdive.com` with their exact requests and responses: default shape, all three models, `answer` present only when requested, `max_results` across its whole range and beyond it, HTTP errors, and client abort. Page content is truncated for readability with the true length kept alongside; nothing else is edited. `evidence/contract-probe.mjs` regenerates the whole file from any key, so none of it has to be taken on trust:

```
SERPDIVE_API_KEY=… node evidence/contract-probe.mjs
```

Three things it settles that are easy to get wrong:

- `max_results` is a **cap between 1 and 10, never a minimum** — the engine returns what it judges relevant, often fewer. Two runs of `max_results: 50` on the same query returned 8 and then 5.
- Out-of-range **numbers** are clamped silently rather than refused, in both directions: `50` becomes 10, and `0` becomes 1 — one result comes back, so `0` is not a way to say "no cap". A value that is **not a number** (`"abc"`, `null`, absent) drops the cap entirely and the default applies. That asymmetry is captured in the file.
- `answer: true` on `krill` is ignored silently, not rejected: the caller still gets a complete result.

**Tests.** Eight focused tests on the current provider routing surface, covering the free default, the opt-in models, the fallback on an unknown model, the `max_results` cap, recency-as-hint, local domain filtering, key redaction on error, and lazy request-time credential resolution. Full suite: 168 passing.

Wired into provider resolution, the auto fallback chain (after Tavily), both curator surfaces, README, and CHANGELOG.

Happy to split this differently if you would rather review the provider and the evidence separately, or to move `evidence/` somewhere else — or drop it from the tree entirely and paste the transcripts here if you would rather not carry them in the repo.

nicobailon:
Thanks for the SERPdive work. I landed it through #167, preserving your commit and adding the small routing-config fix from review so `provider: "serpdive"` and `searchRouting.providers: ["serpdive"]` both work on current `main`.

The changelog credits Eden d'Alexis (`@edendalexis`) for #139. Closing this PR as superseded by #167.

PR #141, opened 2026-07-24
fix(ssrf): add actionable hint for TUN/fake-IP addresses

### LINKED ISSUE #134: SSRF "Blocked internal address" error never mentions the allowRanges opt-in (#101 follow-up)
#101 added `ssrf.allowRanges` in v0.13.0. The runtime error still never points to it, so anyone on a TUN/fake-IP proxy has to find #101 or read source comments to fix it.

## Current behavior

`ssrf-protection.ts`:

```ts
if (ipVersion === 4 && isBlockedIPv4(normalized)) {
  throw new Error(`Blocked internal address for ${hostname}: ${normalized}`);
}
```

On a TUN/fake-IP proxy (Clash/Mihomo/Surge), public hosts resolve into `198.18.0.0/15`, so `fetch_content` and `web_search` URL fetches fail with, e.g.:

```
Blocked internal address for example.com: 198.18.1.22
```

Nothing in the message hints at `allowRanges`.

## Proposal

When the blocked address is in a range fake-IP proxies produce (`198.18.0.0/15`, RFC 2544), append a hint:

```
Blocked internal address for example.com: 198.18.1.22.
Looks like a TUN/fake-IP proxy range (Clash/Mihomo/Surge). To allow it, add to
~/.pi/web-search.json: "ssrf": { "allowRanges": ["198.18.0.0/15"] }
```

The change sits in `assertPublicAddress` next to the existing throw. Default behavior is unchanged.

More opinionated alternative: auto-allow `198.18.0.0/15` by default. It's a non-routable benchmarking range, so the SSRF risk is nil. Happy to PR either.

BenjaminAaron196:
## Summary

- detect blocked addresses in `198.18.0.0/15`, a range commonly used by TUN/fake-IP proxies
- point affected users to the existing `ssrf.allowRanges` configuration
- preserve the secure default and existing errors for all other private/reserved ranges
- add a regression test for the actionable error message

## Testing

- `npm test` (70 tests passed)

Closes #134

### DIFF
--- a/ssrf-protection.ts
+++ b/ssrf-protection.ts
@@ -108,11 +108,19 @@ function assertPublicAddress(address: string, hostname: string, allowRanges: Par
 	// users exempt synthetic ranges produced by TUN/fake-IP proxies (e.g. 198.18/15).
 	if (isInAllowedRange(normalized, ipVersion, allowRanges)) return;
 	if (ipVersion === 4 && isBlockedIPv4(normalized)) {
-		throw new Error(`Blocked internal address for ${hostname}: ${normalized}`);
+		const hint = isFakeIpProxyAddress(normalized)
+			? '. This address is in 198.18.0.0/15, commonly used by TUN/fake-IP proxies. If that matches your setup, configure ssrf.allowRanges with ["198.18.0.0/15"] in web-search.json.'
+			: "";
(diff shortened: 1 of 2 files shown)


nicobailon:
Thanks for this fix. I landed it in #142 with the changelog credit included, and #134 is now closed by the merge.

Issue #136, opened 2026-07-21
Remove `skills/librarian/` from pi-web-access


mcwalrus:
## Problem

The bundled `skills/librarian/ SKILL.md` has no observable effect when this extension is loaded or used. The extension natively exposes `web_search`, `fetch_content` tools to connected models.

## Rationale

If the extension tools are used, then it completely bypasses the skill which has no effect. Personally, I have never observed librarian alongside web-searching tools to have been loaded. Using CLI tools like `git` and `gh  are useful other methods of discoverable search, but are not fleshed out well. Finally, the skill conflicts for any other skills named `librarian` when they fleshed out differently.

## Cost

Tokenised cost computed with `cl100k_base`:

| Component | Bytes (UTF-8) | Tokens | Where it lands |
|---|---:|---:|---|
| `name: librarian` | 15 | 2 | system prompt (per turn) |
| `description:` field | 395 | 65 | system prompt (per turn) |
| frontmatter (incl. `---`) | 419 | 72 | system prompt (per turn) |
| body (markdown) | 7,557 | 1,833 | loaded on demand via `read` |
| **file total** | **7,976** | **1,905** | — 

## Finally

Happily, I use this skill almost everyday. Alongside this critique, I would like to say thank you for the extension as I have found it very useful! I have created a fork of this repo based on my consideration so this is mostly a suggestion.

Regards,

Max

nicobailon:
Thanks for the product critique. The concrete hardcoded clone-path bug in the bundled `librarian` skill was fixed separately in #149, but that intentionally kept the skill installed.

I’m keeping this issue open as the remaining packaging/product decision: either remove the bundled skill and update `package.json`, README, tests, and pack output intentionally, or close this as declined if we want the `/skill:librarian` workflow to remain part of this package. It’s not a one-file cleanup because the skill is advertised through the Pi package metadata and README.

PR #149, opened 2026-07-25
Use reported clone paths in librarian skill

### LINKED ISSUE #23: librarian skill hardcodes /tmp/pi-github-repos, causing brittle path failures
I think pi-web-access is teaching a brittle GitHub repo workflow.

 The main issue seems to be:

 - github-extract.ts defaults clones to /tmp/pi-github-repos
 - skills/librarian/SKILL.md explicitly tells the agent to use paths like:
     - cd /tmp/pi-github-repos/owner/repo
     - cd /tmp/pi-github-repos/owner/repo && git rev-parse HEAD

 This creates a recurring failure mode:

 1. the agent learns /tmp/pi-github-repos/... as the expected clone location
 2. /tmp gets cleared / disappears
 3. later the agent assumes the repo is already there
 4. it immediately fails on cd or read instead of verifying the path first

 In practice, this caused repeated friction for me because the agent kept reusing that assumed path even when it no longer existed. Anyone else experiencing the same thing?

nicobailon:
Fixes #23. Addresses part of #136.

The bundled librarian skill assumed cloned repos lived at `/tmp/pi-github-repos/owner/repo`, but the actual clone root is configurable and session-scoped. This updates the workflow to use the `Repository cloned to:` path returned by `fetch_content`, verify that path still exists before local commands, and refetch when the session-scoped path is gone. It also handles API-only fallback without pretending a local checkout exists.

This keeps the skill installed; fully removing the skill remains a separate product decision for #136.

Validation:
- `node --test test/librarian-skill.test.mjs`
- `npm test` (84 tests)
- `git diff --check origin/main...HEAD`
- fresh reviewer approved

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ All notable changes to this project will be documented in this file.
 - Deferred the heavy content extraction module until the first `fetch_content` or `includeContent` search call, reducing extension startup work. Thanks Kaushik Gopal (@kaushikgopal) for PR #125.
 
 ### Fixed
+- Updated the `librarian` skill to use verified, session-scoped clone paths returned by `fetch_content` instead of assuming `/tmp/pi-github-repos`, and to refetch missing clones. Thanks mcwalrus (@mcwalrus) for #136 and gravewhisper (@gravewhisper) for #23.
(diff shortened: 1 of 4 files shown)


mcwalrus:
Sounds good. I've been happy running a fork so that decision is totally yours.

PR #180, opened 2026-07-28
Add Gemini PDF-to-Markdown conversion with local fallback

jagaliano:
## Summary

This PR upgrades PDF handling from plain local text extraction to structured Markdown conversion with Gemini when a Gemini API key or Cloudflare AI Gateway is configured.

The implementation reuses the repository's existing Gemini transport and credential handling. PDFs are sent as inline `application/pdf` data through a new PDF-specific converter. If Gemini is unavailable or its output is incomplete, blocked, malformed, timed out, or otherwise fails, the existing local `unpdf` extraction remains the fallback.

The original metadata header, output location, filename handling, page count, author, and truncation notice are preserved.

## Motivation

The current `unpdf` path is reliable for text-based PDFs, but it flattens document structure and cannot OCR image-only pages. Gemini document understanding can reconstruct headings, paragraphs, lists, tables, links, footnotes, equations, page boundaries, and scanned text on a best-effort basis.

This approach does not add a new provider, credential, or runtime dependency. It uses the existing Gemini configuration and security controls already exercised elsewhere in the project.

## Implementation

- Added `gemini-pdf-extract.ts` as a separate PDF-specific module.
- Added a generic inline-binary `generateContent` helper to `gemini-api.ts`.
- Kept `video-extract.ts` unchanged; PDF conversion does not share or modify the video upload flow.
- Sends the accepted PDF buffer inline with MIME type `application/pdf`.
- Instructs Gemini to transcribe rather than summarize and to treat the PDF as untrusted source material.
- Preserves the existing `maxPages` behavior, with a default of 100 pages.
- Requires one exact `<!-- Page N -->` marker per requested page and validates marker count and order.
- Rejects non-`STOP` completions such as `MAX_TOKENS`, safety blocks, empty responses, missing candidates, and malformed page structure.
- Removes an enclosing Markdown fence when present and de-duplicates an initial H1 that matches the locally determined title.
- Uses `unpdf` for metadata and as the local extraction fallback.
- Gives PDF.js its own buffer copy because PDF.js may transfer and detach its input buffer.
- Preserves caller cancellation instead of starting fallback work after an abort.
- Separates the PDF download timeout from Gemini processing.

## PDF size configuration

The existing 20 MB behavior remains the default:

```json
{
  "pdf": {
    "maxSizeMB": 20
  }
}
```

Users may optionally raise `pdf.maxSizeMB` to any positive numeric value up to 50. Values above 50 are capped at 50, while invalid, nonnumeric, zero, or negative values fall back to 20.

The 50 MB cap follows Gemini's current [document-processing limit](https://ai.google.dev/gemini-api/docs/generate-content/document-processing). This PR uses inline input for all accepted PDFs; it does not add a Files API upload path.

Size enforcement now checks both:

- the declared `Content-Length`, when present;
- the actual streamed byte count, including responses with a missing or inaccurate length header.

Non-PDF downloads keep their existing 5 MB limit.

## Failure behavior

The conversion path is:

```text
PDF download
  -> read metadata and page count with unpdf
  -> Gemini inline PDF-to-Markdown conversion, when configured
  -> validate completion and page structure
  -> save structured Markdown

If Gemini is unavailable or normal conversion fails:
  -> run the existing local unpdf extraction
  -> save the compatible Markdown result
```

Credential-resolution errors, malformed configuration, and caller cancellation remain visible rather than being hidden by fallback.

## Validation

Automated verification:

- `npm run typecheck` — passed
- `npm test` — **226/226 tests passed**

New and updated tests cover:

- inline PDF request shape and MIME type;
- direct Gemini authentication and Cloudflare AI Gateway routing;
- page limits, marker count, and marker ordering;
- Markdown-fence cleanup and duplicate-title removal;
- truncated, blocked, and empty Gemini responses;
- automatic `unpdf` fallback;
- cancellation without fallback;
- the PDF.js buffer-detachment regression;
- Node 22 compatibility;
- default, custom, invalid, and capped `pdf.maxSizeMB` values;
- exact-limit and oversized streamed PDF responses.

Real Gemini conversion was also verified with three controlled PDFs:

| Example | Result |
|---|---|
| Three-page structured document | Preserved pages 1–3 in order, page markers, headings, lists, and all verification text |
| Financial table and equations | Reconstructed the Markdown table and preserved quantities, the exact total, and LaTeX equations |
| Image-only scanned receipt | OCR succeeded, including accented text, reference number, table structure, and exact total |

The live tests explicitly verified that Gemini produced the Markdown; they did not silently pass through the local fallback.

One expected model limitation was observed: Gemini promoted two H2 headings to H1 in the synthetic multi-page example. No content or page structure was lost, but exact heading hierarchy, layout reconstruction, and OCR remain best-effort model output.

## Documentation

- Updated the PDF section and extraction flow in `README.md`.
- Documented `pdf.maxSizeMB`, its 20 MB default, and 50 MB cap.
- Documented Gemini output limits and local fallback behavior.
- Added the new PDF converter to the source-file table.
- Updated `CHANGELOG.md` under Unreleased.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,13 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Added
+- Added Gemini PDF-to-Markdown conversion with inline document input, page-structure validation, OCR/layout preservation on a best-effort basis, and automatic local `unpdf` fallback for incomplete or failed model output.
+- Added optional `pdf.maxSizeMB` configuration with the existing 20MB default, a 50MB cap, early `Content-Length` rejection, and streamed byte enforcement for missing or inaccurate response lengths.
+
+### Changed
+- Separated the PDF download timeout from Gemini processing and preserved caller cancellation without changing local video extraction.
+
 ## [0.15.0] - 2026-07-28
 
 ### Added
--- a/README.md
+++ b/README.md
@@ -199,7 +199,9 @@ Requires `ffmpeg` (and `yt-dlp` for YouTube). Timestamps accept `H:MM:SS`, `MM:S
 
 ### PDFs
 
-PDF URLs are extracted as text and saved to `~/Downloads/` as markdown. The agent can then `read` specific sections without loading the full document into context. Text-based extraction only — no OCR.
+PDF URLs are converted to structured Markdown and saved to `~/Downloads/`. When the Gemini API or a configured Cloudflare AI Gateway is available, the converter sends the PDF inline to Gemini, preserving page markers, headings, lists, tables, links, footnotes, equations, and scanned text on a best-effort basis. Incomplete, blocked, malformed, timed-out, or failed Gemini output automatically falls back to local `unpdf` text extraction.
+
+The download limit remains 20MB by default. Set `pdf.maxSizeMB` to raise it up to Gemini's [documented 50MB PDF input limit](https://ai.google.dev/gemini-api/docs/generate-content/document-processing). Both `Content-Length` and the bytes actually received are enforced. The default extraction limit is 100 pages; longer documents include a truncation notice.
 
 ### Blocked pages
 
@@ -215,7 +217,7 @@ fetch_content(url)
   → Video file?  Gemini API (Files API) → Gemini Web (if browser cookies enabled)
   → GitHub URL?  Clone repo, return file contents + local path
   → YouTube URL? Gemini Web (if browser cookies enabled) → Gemini API → Perplexity
-  → HTTP fetch → PDF? Extract text, save to ~/Downloads/
+  → HTTP fetch → PDF? Gemini Markdown conversion → unpdf fallback → save to ~/Downloads/
                → HTML? Readability → RSC parser → Firecrawl (if configured, cache-only by default) → Jina Reader → TinyFish → Parallel → Gemini fallback
                → Text/JSON/Markdown? Return directly
 ```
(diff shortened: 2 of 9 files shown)


nicobailon:
Thanks for the detailed PDF conversion work. I reviewed this as part of the backlog pass and I’m going to leave it open for now rather than merge it in its current form.

The branch is still a draft, maintainers can’t modify it, and it overlaps with two decisions/changes that have moved since it was opened:

- #183 has now landed, so the PDF default output directory is `tmpdir()/pi-web-pdf` rather than `~/Downloads`. This PR still keeps the old Downloads default in its rewritten `pdf-extract.ts`.
- #181 needs an owner decision on the default Gemini model and Gemini Web fallback policy before this new Gemini PDF path should become the source of truth. The PDF path inherits the shared Gemini API default, so it should be updated together with that decision.

Recommended next step: rebase after current `main`, keep the temp-dir output behavior from #183, and align the PDF Gemini model choice with whatever policy we choose for #181. Once it’s no longer draft and those two points are resolved, it should be reviewed as a larger feature PR with fresh focused gates.

nicobailon:
Follow-up: #181 is now implemented on `main` in 146ec74.

For the next revision of this draft, please rebase on current `main` and keep both of these current behaviors:

- PDF markdown defaults to the temp `pi-web-pdf` directory from #183.
- Gemini API defaults use `gemini-3.6-flash`; Gemini Web remains separate and uses its browser-header-specific `gemini-3.1-pro` default.

After that, the PDF-to-Markdown conversion can be reviewed as a focused feature PR on top of the current model/output policy.

nicobailon:
Thanks again for the Gemini PDF-to-Markdown work here.

I landed this as a clean maintainer replacement in #188, now merged at d16353c. The replacement preserves the current `main` policies that changed after this draft was opened: PDF markdown still defaults to the temp `pi-web-pdf` directory, and Gemini API uses `gemini-3.6-flash` while Gemini Web remains separate.

I kept visible credit for José Antonio Galiano Sandoval (`@jagaliano`) / PR #180 in the changelog and included you as co-author on the replacement commit.

PR #190, opened 2026-07-31
fix: googlesearch tool gemma model compatibility

pricewa:
now compatible with gemini AND gemma-4-31b-it, gemma-4-26b-a4b-it


nicobailon:
Thanks for the PR. I’m going to close this one because the current change would make the REST request diverge from Gemini’s documented API shape.

The code here sends `models/*:generateContent` directly over REST, where Google’s docs show the Search grounding tool as `tools: [{ "google_search": {} }]`. The `googleSearch` casing is used by the JS SDK object shape, but changing this REST payload to `googlesearch` is not supported by the contract we’re calling and would risk breaking Gemini API search grounding.

PR #196, opened 2026-08-02
Add explicit-only xAI (Grok) search provider

join3r:
Adds `provider: "xai"`, backed by xAI's Agent Tools API — the hosted `web_search` tool on `https://api.x.ai/v1/responses`.

## Why

`openai-search.ts` already lets a Codex subscription pay for its own searches by resolving auth through Pi's model registry. The same is now true for xAI: a user signed in with SuperGrok or X Premium gets web search with no key to configure. `xaiApiKey` / `XAI_API_KEY` remain as the pay-as-you-go fallback, via the usual `credential-source` path.

## Explicit-only, on purpose

It follows AnySearch: excluded from `auto` fallback and from `provider: "all"`, selectable with `provider: "xai"`, as the configured provider, or in `searchRouting`.

The reason is cost shape rather than taste. The search runs inside Grok's own inference, and one question typically fans out to ~11 `web_search` tool calls, billed per call on top of tokens and drawn from the same allowance the account uses for chatting. A provider that can quietly spend the user's chat quota shouldn't be reachable by `auto`.

## Empirically grounded

All of this was probed against a live SuperGrok account rather than inferred from docs:

- The request body is deliberately the minimal `{ model, input, tools }` shape that was verified to work. `stream`, `include`, `tool_choice` and `parallel_tool_calls` — which the OpenAI backend sends — were **not** verified against xAI, and a 400 from an unsupported field would cost the user the whole search. `web_search_call.action.sources` comes back without asking for it via `include`.
- Recency, domain and result-count options are folded into the prompt text rather than sent as tool parameters, for the same reason.
- Sources are read from `url_citation` annotations on the answer **and** from each `web_search_call`'s own sources. This API has **no top-level `citations` array** — reading one (as at least one other Pi search extension does) silently yields answers with no sources. There's a fixture asserting exactly this.
- xAI's older Live Search (`search_parameters` on `/v1/chat/completions`) is deprecated and now answers **HTTP 410** with a pointer to the Agent Tools API. Worth knowing before anyone adds it.

## Tests

`test/xai-provider.test.mjs`, 8 cases in the style of `anysearch-provider.test.mjs`: request shape, both citation shapes, prompt-folded filters, explicit routing, exclusion from `auto`, exclusion from `all`, credential redaction, and the no-answer-no-sources failure. Full suite is 278 passing, `npm run typecheck` clean.

README updated: provider lists, the `all` exclusion note, the credential-source and env-var lists, the file table, and a new "xAI (Grok)" section.


nicobailon:
Thanks @join3r. I landed this via maintainer PR #209 so the xAI provider could be rebased cleanly and include the missing curator support. Your contribution is preserved in the landed commit authorship and changelog credit, so I am closing this PR as superseded.

nicobailon:
Superseded by merged maintainer PR #209.

PR #218, opened 2026-08-07
fix: add package-lock

dougEfresh:
Pretty please add package-lock.json

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
 node_modules/
 bun.lock
-package-lock.json
 .pi-subagents/
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,5499 @@
+{
+  "name": "pi-web-access",
+  "version": "0.18.0",
+  "lockfileVersion": 3,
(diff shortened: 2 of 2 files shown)


nicobailon:
Thanks for the PR. I am going to close this because this extension repo intentionally does not track npm lockfiles. The current changelog records that the committed `package-lock.json` was removed and future lockfiles are ignored, and `.gitignore` still carries that policy on `main`. So this would reverse the current package policy rather than fix a repo bug.

dougEfresh:
What is the technical justification?

This is your project , your decision.  You provide great extensions and I am grateful 

But you really put users at significant risk.

nicobailon:
I took another pass after your follow-up. The lockfile idea is valid, but the original PR now makes npm ci install current Pi peer packages and npm run typecheck fails unless the extension also accepts the current ProviderHeaders type. I opened #223 as a replacement that preserves your package-lock contribution, adds changelog credit for #218, and includes the small compatibility fix needed for current main. Thanks again for pushing on this.

PR #238, opened 2026-08-10
fix: externalize fetched content storage

### LINKED ISSUE #236: Full fetched content stored in session JSONL causes 100+ MB sessions and repeated Pi freezes
## Summary

Long research sessions using `web_search` eventually make Pi repeatedly freeze at `Working...`.

`pi-web-access` persists complete fetched-page payloads as `custom:web-search-results` entries using `pi.appendEntry()`. Because Pi sessions are append-only, these payloads permanently inflate the JSONL session—even after their one-hour in-memory TTL expires and after conversation compaction.

## Observed behavior

Two research sessions grew to:

- 165.6 MiB, including 89 `web-search-results` entries
- 164.9 MiB, including 264 `web-search-results` entries

In the first session, those custom entries contained approximately 77.8 MiB of JSON characters. Individual entries reached 11 MiB.

When the problem occurs:

- Pi remains at `Working...`
- the Pi process consumes approximately 100% CPU
- RSS grows to approximately 1.7–2.0 GiB
- there is no active network connection
- session output stops advancing
- stack sampling shows repeated local array/string processing
- restarting or compacting only helps temporarily

The freeze can occur at low model-context usage. One occurrence happened at 23.8% of a 272k context window, so this is distinct from a normal context-window overflow.

After replacing only the large `web-search-results` payloads with small metadata placeholders:

- current session: 165.6 MiB → 38.6 MiB
- previous session: 164.9 MiB → 21.0 MiB
- Pi resumed the exact same session successfully
- startup RSS dropped to approximately 356 MiB
- the session became responsive again

## Why compaction does not fix it

Pi compaction reduces the live LLM context but intentionally retains full history in the append-only JSONL file.

The large entries remain in the session and are traversed again when the session is loaded. `restoreFromSession()` also walks the branch and restores recent `web-search-results` payloads into memory.

The one-hour TTL prevents old results from entering the in-memory result map, but does not remove their full payloads from the JSONL session.

## Relevant implementation

`startBackgroundFetch()` stores full extracted content:

```ts
const data: StoredSearchData = {
  id: fetchId,
  type: "fetch",
  timestamp: Date.now(),
  urls: stripThumbnails(fetched),
};

storeResult(fetchId, data);
pi.appendEntry("web-search-results", data);
```

`restoreFromSession()` then traverses custom entries:

```ts
for (const entry of ctx.sessionManager.getBranch()) {
  if (entry.type === "custom" && entry.customType === "web-search-results") {
    // ...
  }
}
```

## Reproduction

1. Start a persistent Pi session with `pi-web-access`.
2. Perform repeated broad `web_search` calls that background-fetch dozens of result URLs.
3. Continue researching for an extended period.
4. Observe repeated `web-search-content-ready` messages.
5. Inspect the session JSONL:
   - many `custom:web-search-results` entries contain full extracted pages
   - the file grows past 100 MiB
6. Continue the agent or restart/resume the session.
7. Pi eventually becomes CPU-bound at `Working...`.

## Expected behavior

Large fetched content should not be embedded directly in the Pi session file.

The session should retain only bounded metadata needed to locate the stored response, while full page content should live in an external cache and be loaded lazily by `get_search_content`.

## Suggested approach

- Store full response payloads under a private cache directory.
- Persist only response ID, type, timestamp, metadata, and cache path/key in the Pi custom entry.
- Load content lazily when `get_search_content` is called.
- Apply TTL cleanup to external cache files.
- Place a conservative maximum size on any custom entry written to the session.
- Gracefully support existing inline session entries during migration.
- Optionally provide a command to externalize legacy inline payloads.

## Acceptance criteria

- Repeated broad searches do not cause session JSONL size to scale with full fetched-page content.
- Resuming a search-heavy session does not require loading all fetched content.
- `get_search_content` continues to retrieve stored content on demand.
- Expired cached content is removed without rewriting the Pi session.
- Tests cover hundreds of fetched pages and session resume behavior.
- Custom session entries remain bounded.

## Environment

- pi-web-access: 0.19.0
- Pi: 0.84.1
- Node.js: v24.13.0
- macOS: 15.6.1

I also checked pi-web-access 0.20.0; `storage.ts` and this persistence behavior appear unchanged.

## Related issue

#112 addressed unbounded content returned by `get_search_content`, but this is a different problem: full content is persisted in the session JSONL even when it is not returned to the model.

The affected sessions contain private research data, so I cannot attach them, but the aggregate measurements above are from the actual recovery.

nicobailon:
Fixes #236.

This moves full fetched-page bodies out of Pi session custom entries and into a private cache. Session entries now keep bounded metadata plus cache references, while `get_search_content` still loads the content on demand. Legacy inline session entries remain readable.

Validation:
- `npm test -- test/fetch-cache-storage.test.mjs test/get-search-content.test.mjs test/fetch-answer-storage.test.mjs`
- `npm run typecheck`
- `git diff --check origin/main...HEAD`

Credit: thanks Igor Samokhovets (`@samohovets`) for the report and recovery measurements in #236.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ All notable changes to this project will be documented in this file.
 - Added per-tool and per-command registration gates plus image and PDF extraction gates. Thanks @jaudiger for issue #234.
 
 ### Fixed
+- Store full fetched content in an external cache instead of embedding it in session JSONL entries, preventing large search-heavy sessions from ballooning on restore. Thanks Igor Samokhovets (`@samohovets`) for issue #236.
 - Document `get_search_content` parameter constraints in the tool schema. Thanks `@iwangjie` for PR #233.
 
 ## [0.20.0] - 2026-08-10
--- a/README.md
+++ b/README.md
@@ -148,7 +148,7 @@ fetch_content({ url: "https://example.com/diagram.png" })
 
 ### get_search_content
 
-Retrieve stored content from previous searches or fetches. Fetched URL content is stored in full, including when `fetch_content` uses answer mode. Use `findText` to locate bounded matching passages without paging through a large page, or use `offset` and `limit` to retrieve slices intentionally.
+Retrieve stored content from previous searches or fetches. Fetched URL content is stored in full in a private `web-search-cache` directory under the Pi config directory, not in the session JSONL. This includes `fetch_content` answer mode, which stores the original page content. Cache entries use the same one-hour lifetime as in-session result IDs. Use `findText` to locate bounded matching passages without paging through a large page, or use `offset` and `limit` to retrieve slices intentionally.
 
 ```typescript
 get_search_content({ responseId: "abc123", urlIndex: 0 })
--- a/index.ts
+++ b/index.ts
@@ -17,6 +17,7 @@ import {
 	getAllResults,
 	getResult,
 	restoreFromSession,
+	storeFetchedContentResult,
 	storeResult,
 	type QueryResultData,
 	type StoredSearchData,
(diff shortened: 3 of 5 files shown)


samohovets:
Thanks @nicobailon! Will test it out

PR #249, opened 2026-08-13
fix: preserve provider summary completion routing

limwa:
Always route summary completions through the model registry.

Fixes an issue with custom providers that wrap provider streams, namely by intercepting requests and modifying headers, or rewriting model IDs. Summary generation with these providers fails, as the `complete` function from the compat model doesn't call the provider's functions.


nicobailon:
Closed as superseded by #252, which preserves the provider summary completion routing fix, keeps contributor credit in the changelog, and passed full local validation plus the Greptile gate.

PR #276, opened 2026-08-19
feat: configurable Pi provider priority for OpenAI web search

hank-warren:
`openai-search.ts` hardcodes which Pi providers it resolves auth from:

```ts
const SEARCH_PROVIDERS = ["openai-codex", "openai"] as const;
```

Since `openai-codex` is tried first and wins whenever it's signed in, there's no way to route search spend anywhere else — e.g. to a metered `openai` API key instead of the Codex subscription's weekly limit, or to a second Codex account slot registered via `pi.registerProvider()` (my case: I keep a separate account for background/automation traffic and wanted `web_search` billed there rather than against my main subscription).

Everything downstream of the list already generalizes: auth resolves through `getApiKeyAndHeaders`, the endpoint is chosen by inspecting the token (`isCodexJwt`), and `chatgpt-account-id` is extracted from the JWT itself. The provider list is the only hardcoded piece, so this PR just makes it configurable:

```json
{ "openaiSearchProviders": ["openai-codex-work", "openai-codex"] }
```

Purely additive:

- unset → exactly today's behavior (`["openai-codex", "openai"]`)
- entries that aren't registered or signed in are skipped by the existing loop
- `[]` skips Pi credentials entirely, so the `openaiApiKey` / `OPENAI_API_KEY` fallback applies (useful for "signed into Codex, but bill searches to my API key")
- anything that isn't an array of non-empty strings fails with the same error style as the other config fields

Added two tests in `test/search-providers.test.mjs` covering priority order (including skipping an unregistered entry) and config validation. `npx tsc` and `node --test` pass, plus a README note and a changelog line.


hank-warren:
Fair catch on the context-gated validation — malformed `openaiSearchProviders` could silently fall through to the API-key path when no extension context was passed. Fixed in 898ad77: the config is now validated unconditionally in both `resolveOpenAIAuth` and `isOpenAISearchAvailable`, matching how `openaiResponsesUrl` is handled. Extended the invalid-config test to cover the context-free path and assert no request is ever sent with the fallback key.

PR #251, opened 2026-08-14
Constrain numeric tool parameters

jaudiger:
Fixes #250

I ran into this while retrieving fetched page content.

This change:
- Adds schema bounds for numeric tool parameters.
- Improves get_search_content errors with received values, valid ranges, and retry guidance.
- Extends the existing focused tests without adding new test files.

PR #295, opened 2026-08-23
fix(get_search_content): tolerate bridge defaults with findText

ZacharyQin:
## Summary

Make `get_search_content` tolerate default fields injected by tool-call bridges when using `findText`.

## Problem

When an agent requests text matching from stored search or fetch content, the tool-call bridge may include:

- `offset: 0`
- `limit: 1`, `10000`, or `20000`
- `query: ""`
- `url: ""`

The current implementation rejects `findText` whenever `offset` or `limit` is present. Empty selectors can also take precedence over `queryIndex` or `urlIndex`.

As a result, valid stored content cannot be searched and the agent may repeatedly retry the same invalid call.

## Root Cause

The tool schema marks `offset`, `limit`, `query`, and `url` as optional, but the caller may serialize them as empty or default values. The runtime treats those fields as intentional input instead of normalizing bridge defaults.

## Changes

- Normalize empty `query` and `url` to `undefined`.
- Give `findText` precedence over `offset` and `limit`.
- Ignore `offset` and `limit` when `findText` is present.
- Preserve existing pagination behavior when `findText` is absent.
- Update the tool schema descriptions to document this behavior.
- Add regression coverage for polluted search and fetch calls.

## Compatibility

- Existing `get_search_content` slice requests continue to use `offset` and `limit`.
- Existing `queryIndex` and `urlIndex` requests continue to work when the corresponding selector string is empty.
- Calls that provide `findText` together with `offset` or `limit` now return matching passages instead of an incompatibility error.

## Validation

Focused regression test:

```bash
node --experimental-strip-types --test test/get-search-content.test.mjs
```

Result:

```text
7 pass 0 fail
```

Type check:

```bash
npm run typecheck
```

Result: passed.

Also verified the registered tool with local search and fetch fixtures:

- `findText` search result lookup: passed
- `findText` fetched-content lookup: passed
- polluted empty-selector and pagination defaults: passed
- normal paginated slice lookup: passed

The full local suite was also run with `NODE_OPTIONS=--experimental-strip-types`: `504 pass / 16 fail`. The remaining failures are environment-dependent existing tests involving browser-cookie fixtures, temporary process fixtures, package-install network access, and Node child-process loading under the local Node `22.17.0` runtime.

## Risks

`offset` and `limit` are ignored when `findText` is present. This avoids ambiguity because matching mode returns bounded matching passages rather than a positional content slice.

## References

- PR #233: document `get_search_content` parameter constraints
- PR #251: numeric parameter validation
- PR #205: fix the related `fetch (no URL)` display issue

### DIFF
--- a/get-search-content-params.ts
+++ b/get-search-content-params.ts
@@ -1,0 +1,26 @@
+export interface GetSearchContentParams {
+	responseId: string;
+	query?: string;
+	queryIndex?: number;
+	url?: string;
+	urlIndex?: number;
+	offset?: number;
+	limit?: number;
+	findText?: string | string[];
+	findMode?: string;
+}
+
+export function normalizeGetSearchContentParams(params: GetSearchContentParams): GetSearchContentParams {
+	// Tool bridges may serialize optional selectors and slice defaults even when unset.
+	const normalized = { ...params };
+
+	if (normalized.query?.trim() === "") delete normalized.query;
+	if (normalized.url?.trim() === "") delete normalized.url;
+
+	if (normalized.findText !== undefined) {
+		delete normalized.offset;
+		delete normalized.limit;
+	}
+
+	return normalized;
+}
--- a/index.ts
+++ b/index.ts
@@ -4,6 +4,7 @@ import { Type } from "typebox";
 import { StringEnum, type ImageContent, type TextContent } from "@earendil-works/pi-ai/compat";
 import type { ExtractedContent, ExtractOptions } from "./extract.ts";
 import { normalizeFetchContentParams } from "./fetch-params.ts";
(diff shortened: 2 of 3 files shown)


nicobailon:
Thanks for the fix. I landed this through owner-branch replacement PR #303 so I could keep the repo-local changelog/credit gates and avoid pushing to your fork. The `get_search_content` bridge-default behavior is preserved, and `CHANGELOG.md` keeps credit: Thanks to [@ZacharyQin](https://github.com/ZacharyQin) for PR #295.

nicobailon:
Closing as superseded by merged replacement PR #303. Thanks again for the contribution.

PR #301, opened 2026-08-24
feat: authenticate Gemini generate-content via Application Default Credentials (Vertex AI)

smazurov:
## Summary

Adds `geminiAuth: "adc"` so Gemini `generate-content` calls — **web search, URL context, and PDF/inline-data extraction** — can authenticate with **Google Application Default Credentials** instead of a `GEMINI_API_KEY`.

Requests now target the Vertex AI endpoint (`aiplatform.googleapis.com`) with an OAuth **bearer token minted from the ADC file** (`authorized_user` refresh-token exchange or `service_account` JWT assertion), with project/location from `geminiProject`/`geminiLocation` (falling back to `GOOGLE_CLOUD_PROJECT`/`GOOGLE_CLOUD_LOCATION`/`GCLOUD_PROJECT`). Tokens are **cached and refreshed** from expiry and **redacted from all errors**.

## Precedence (backward compatible)

- An explicit `GEMINI_API_KEY` / `geminiApiKey` still **wins** over ADC.
- An explicit `GOOGLE_GEMINI_BASE_URL` / `geminiBaseUrl` (gateway/relay/proxy) still **wins** over ADC.
- Default behavior (no `geminiAuth`) is unchanged.

## Scope

ADC covers paths routed through the shared `fetchGeminiApi` transport: search, URL context, and inline-data (PDF/image) generation. **YouTube and local video analysis** keep requiring `GEMINI_API_KEY` because they use the Gemini Files API, which has a different Vertex flow; they fall back to Gemini Web as before.

## Tests

New `test/gemini-adc-auth.test.mjs` (3 tests):
- Vertex URL construction + `Authorization: Bearer` header, and no `x-goog-api-key` in ADC mode
- Availability gating (`geminiAuth`, project/location, ADC file presence)
- Access token never leaks in error messages

Validated:
- `node --test` full suite: **524/524 pass** (existing `gemini-api-transport` and `gemini-pdf-extract` tests unaffected)
- Live end-to-end against a real ADC-backed Vertex project: `web_search` provider `gemini` returns grounded results with citations

## Files
- `gemini-adc.ts` (new): ADC discovery, token minting/caching, redaction
- `gemini-api.ts`: ADC-aware base URL, auth headers, redaction, availability
- `gemini-search.ts`, `gemini-url-context.ts`: availability gates for ADC mode
- `README.md`, `CHANGELOG.md`: docs


smazurov:
Addressed both Greptile findings, but red-teamed them first — one is fixed as suggested, the other fixed more narrowly than suggested.

## Conflict
Resolved by merging `main` into the branch (`3a69fbf`). Only `CHANGELOG.md` conflicted (two `### Added` sections under `[Unreleased]`); both entries preserved. PR is now `MERGEABLE`.

## P2 — ADC overstates YouTube availability (agreed, fixed)
`isGeminiApiAvailable()` now returns true with ADC-only credentials, but YouTube/local video route through the Gemini **Files API**, which needs an API key and can't use ADC. Result was a guaranteed-failing `queryGeminiApiWithVideo` call plus a misleading "Gemini API not configured" error on every YouTube extract.

Fix: added `isGeminiApiAvailableWithVideo()` (API key or Cloudflare gateway only, no ADC) and switched `youtube-extract.ts` to it. Search/PDF/URL-context keep the ADC-inclusive predicate, so no behavior changed for the paths ADC actually supports.

## P1 — ADC token-exchange failures bypass the PDF credential boundary (partial agree, narrower fix)
The underlying concern is valid: a rejected ADC credential made explicit Gemini PDF extraction silently fall back to local `unpdf`. But the suggested blanket fix — classify **all** token-exchange failures as `CredentialResolutionError` — is too broad:

- A network error / timeout / 5xx from `oauth2.googleapis.com` is transient; making it fatal would break the Gemini Web fallback for search and the local fallback for PDF that this PR explicitly advertises.
- It would also suppress Gemini Web fallback in `gemini-search.ts`, where `CredentialResolutionError` is rethrown by design.

Fix (`1eaa7a2`): only OAuth **4xx rejections** (revoked/bad refresh token, bad client, insufficient scopes — i.e. the credentials themselves were refused) now throw `CredentialResolutionError` (new category `oauth-credential-rejected`). This hooks into the codebase's existing credential boundary: explicit Gemini PDF extraction rethrows it instead of downgrading, and search/URL-context surface it via the existing `CredentialResolutionError` handling. Transient failures still fall back normally.

## Tests
Added regression tests for both: ADC OAuth rejection → `CredentialResolutionError` + 5xx → plain `Error`; ADC-only availability excludes the Files-API path while API-key restores it. Full suite: 541/541 pass, `npm run typecheck` clean.

smazurov:
Valid catch — retracting the blanket 4xx classifier I introduced in `1eaa7a2`. An OAuth HTTP **429** is transient/retryable and must not be treated as a credential rejection, or rate-limited Gemini PDF extraction would now hard-fail instead of falling back to local `unpdf`.

Fixed in `04f7042`: `throwOnTokenExchangeFailure` now only maps the actual credential-rejection statuses — **400/401/403** (`invalid_client`/`invalid_grant`/`invalid_scope`, denied) — to `CredentialResolutionError`. **429** rate limits, 404, all 5xx, and network errors stay plain `Error`, preserving the Gemini Web / local PDF fallbacks for everything transient.

The regression test now asserts the full split (400/401/403 → `CredentialResolutionError`; 429/404/500/502/503 → plain `Error`). Full suite 541/541, `tsc` clean.

nicobailon:
Thanks for the ADC implementation. I landed this through owner-branch replacement PR #304 so I could keep the repo-local changelog/credit gates and avoid pushing to your fork. The Gemini ADC / Vertex AI behavior is preserved, and `CHANGELOG.md` keeps credit: Thanks to [@smazurov](https://github.com/smazurov) for PR #301.

nicobailon:
Closing as superseded by merged replacement PR #304. Thanks again for the contribution.

Issue #104, opened 2026-06-25
Feature: Support configurable HTTP proxy (and timeout) for fetch_content & search APIs


mystery4f:
## Problem

Users behind restricted networks (e.g. mainland China) cannot use `fetch_content` to retrieve foreign pages, and there is no way to route any of the plugin's outbound requests through an HTTP proxy.

`fetch_content` proxies every target URL through **Jina Reader** (`https://r.jina.ai/`), which is unreachable without a proxy in restricted regions:

```
$ fetch_content("https://www.google.com")
→ Error: fetch failed

$ fetch_content("https://www.baidu.com")
→ works (returns valid markdown)
```

The domestic-vs-foreign split confirms this is a network-reachability issue, not a bug in the extraction logic.

Looking at the source, **all** outbound network calls are bare `fetch()` with no proxy option, across every module:

```ts
// extract.ts — Jina Reader fetch (the URL-extraction backend)
const JINA_READER_BASE = "https://r.jina.ai/";
const res = await fetch(jinaUrl, { headers: { ... } });

// exa.ts, brave.ts, perplexity.ts, gemini-*.ts, parallel.ts, tavily.ts, openai-search.ts
const response = await fetch(EXA_SEARCH_URL, { ... });
```

The `fetch_content` tool parameters don't even declare a `proxy` (or configurable `timeoutMs`) field:

```ts
// fetch-params.ts
export interface FetchContentParams {
  url?: unknown; urls?: unknown;
  forceClone?: unknown; prompt?: unknown;
  timestamp?: unknown; frames?: unknown; model?: unknown;
  // ← no proxy, no timeoutMs
}
```

## Why it matters

In regions where `r.jina.ai`, `api.exa.ai`, Gemini, etc. are blocked, the documented fallbacks (Gemini API key / Parallel API key / Chrome login) all hit the same unreachable endpoints, so they don't actually help. The only workaround today is an OS-level transparent proxy (TUN mode), which is invasive and not always available.

The predecessor tool (`web_fetch` from `pi-smart-fetch`) solved this cleanly with a per-call `proxy` parameter, which worked great. It would be fantastic to bring the same capability here.

### 1. Proxy support — any one (or more) of:

- A global `httpProxy` field in `web-search.json`, applied to every outbound `fetch` (lowest friction)
- A per-call `proxy` parameter on `fetch_content` / `web_search`
[...]

PR #307, opened 2026-08-25
feat: curl-backed proxy transport for web_search/source_check/fetch_content

mystery4f:
## Summary

Adds an optional `proxy` parameter to `web_search`, `source_check`, and `fetch_content`. When provided, all outbound HTTP(S) requests are routed through `curl` instead of Node built-in fetch.

This solves the core reachability problem for users behind restricted networks (e.g. mainland China) where `undici` ignores `HTTP(S)_PROXY` env vars and `ProxyAgent` fails TLS handshakes against common HTTP proxies (`ERR_SSL_WRONG_VERSION_NUMBER`).

## Changes

- **utils.ts** (+272): curl-backed transport layer, global fetch wrapper (`installGlobalProxyFetch`), `NO_PROXY` matching, per-call proxy state management
- **fetch-params.ts** (+24): `normalizeProxy()` parameter validation
- **index.ts** (+15): Three tools wired with `proxy` param + `setActiveProxy` calls
- **ssrf-protection.ts** (+7): Skip local DNS preflight when proxy is active (DNS resolves on proxy side)
- **extract.ts** (+2): `ExtractOptions.proxy` field
- **README.md** (+15): Proxy documentation
- **CHANGELOG.md** (+1): Entry

## Relation to existing work

Complements PR #153 (merged) which added `ssrf.trustEnvProxy` to skip DNS preflight for env-proxied hosts. That PR addresses DNS-level SSRF false positives but does **not** provide actual proxy transport — Node fetch still ignores proxy env vars.

This PR provides the transport layer: when `proxy` is set (per-call or via `web-search.json` global config), all HTTP(S) requests go through `curl -x <proxy>`, which handles TLS correctly against proxies that undici cannot.

## Configuration

Per-call parameter:
```typescript
web_search({ query: "...", proxy: "http://host:port" })
fetch_content({ url: "...", proxy: "http://host:port" })
```

Global config in `~/.pi/web-search.json`:
```json
{ "proxy": "http://mcr:4444" }
```

Empty string `""` forces direct connection even when global proxy is set.

## Backward compatibility

All changes are additive. `proxy` is Optional everywhere. Without the parameter, behavior is identical to before.

## Fixes

Relates to #104 (original feature request, closed by #153 but proxy transport was not included in that PR).


mystery4f:
Addressed all 3 greptile-apps review findings:

**1. Proxy state crosses tool calls (P1 Security)** - Fixed.
Replaced module-level perCallProxy with AsyncLocalStorage to scope proxy per async context. Background fetches now inherit their originating tool calls proxy and cannot be overwritten by concurrent tool calls.

**2. Proxy bypasses SSRF address checks (P1 Security)** - Fixed.
DNS preflight skip now requires trustEnvProxy: true (the existing opt-in from PR 153). Caller-supplied proxies no longer unconditionally bypass the public-address validation boundary.

**3. Redirect error mode is ignored (P1)** - Fixed.
redirect: error now correctly prevents curl from following redirects, matching native fetch behavior for protected callers (Gemini, Datalab).

All changes verified with tsc --noEmit (zero errors).

mystery4f:
Addressed the new SSRF finding:

**Caller proxies inherit SSRF trust** - Fixed.
Removed the `getActiveProxy()` early-return entirely. The existing `shouldTrustEnvProxy()` already handles env proxy trust correctly. Caller-supplied proxies now always go through DNS validation, preventing arbitrary proxy parameters from inheriting the `trustEnvProxy` setting.

Verified with `tsc --noEmit` (zero errors). Please re-review.

mystery4f:
@greptile-apps The `getActiveProxy()` early-return in ssrf-protection.ts has been removed. Only `shouldTrustEnvProxy()` (env proxy trust) remains. Please re-review the latest commit ec3fcb8.

mystery4f:
Fixed "Queued fetches inherit wrong proxy":

Introduced `ProxiedRequestInit.__proxy` — callers attach the proxy directly on the fetch init object. The global wrapper prefers `__proxy` over `AsyncLocalStorage`, so pLimit-queued `extractContent` calls use the correct proxy even when the async context shifts.

Verified with `tsc --noEmit` (zero errors). Please re-review.

PR #308, opened 2026-08-25
fix: point 404/410 fetch failures at search instead of provider config

Daniishkhan:
## Summary

When `fetch_content` gets a definitive 404 or 410 from the origin server, the error currently ends with the "Fallback options" checklist (configure Firecrawl, TinyFish, Kagi, etc.). No extraction provider can retrieve a page the origin says is gone, so the checklist sends users down the wrong path — I hit this fetching an Apollo docs page that had been renamed, and spent a while looking at provider config before realizing the URL itself was dead.

This replaces the checklist for 404/410 responses with a short note that the page likely moved and a pointer to `web_search` to find the current URL. All other failures keep the existing checklist unchanged.

Before:

```
HTTP 404: Not Found

Fallback options:
  • Set firecrawlBaseUrl in ~/.pi/web-search.json to a self-hosted Firecrawl instance
  • Set tinyfishApiKey in ~/.pi/web-search.json or TINYFISH_API_KEY
  ...
```

After:

```
HTTP 404: Not Found

The origin server says this page does not exist (HTTP 404), so extraction providers cannot retrieve it.
The page may have moved or been renamed. Use web_search to find the current URL, then retry fetch_content with it.
```

The fallback chain itself is untouched — a provider that can still serve the page (e.g. a cached copy) returns successfully before this message is built, so this only changes the all-providers-failed guidance text.

## Changes

- `extract.ts`: short-circuit the guidance builder for HTTP 404/410 with not-found-specific guidance.
- `test/fetch-not-found-guidance.test.mjs`: covers 404, 410, and asserts transient errors (500) still get the full checklist.

## Test plan

- `node --test test/fetch-not-found-guidance.test.mjs` — 3/3 pass
- `npm test` — 559/560 pass; the one failure (`chrome-cookie-extraction` › "failed password lookups are retried instead of cached") also fails for me on unmodified `main`, so it looks environment-related and unrelated to this change
- `npm run typecheck` — clean

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ All notable changes to this project will be documented in this file.
 ### Fixed
 - Added Defuddle as a local fallback when Readability and RSC extraction cannot recover useful HTML content. Thanks to [@tobru](https://github.com/tobru) for issue #300.
 - Kept Gemini ADC config parse errors, GitHub PR/issue REST fallback failures, and Defuddle fallback failures visible instead of silently downgrading them.
+- Pointed definitive `fetch_content` 404/410 failures to search guidance instead of provider configuration. Thanks to [@Daniishkhan](https://github.com/Daniishkhan) for PR #308.
 - Improved Gemini Web browser-cookie diagnostics so `/google-account` shows sanitized attempted browser/profile entries and distinguishes missing required cookies, password-store access, and decryption failures. Thanks to [@lmilojevicc](https://github.com/lmilojevicc) for issue #296.
 - Made `get_search_content` tolerate bridge defaults when `findText` is supplied, while preserving ordinary pagination. Thanks to [@ZacharyQin](https://github.com/ZacharyQin) for PR #295.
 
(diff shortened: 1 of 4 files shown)


Daniishkhan:
Good catch — febb844 resolves the configured public tool names (`toolNames.webSearch` / `toolNames.fetchContent`) for the guidance text instead of hardcoding `web_search`/`fetch_content`, with a fallback to the defaults when the config is missing or malformed. Also fixed the same hardcoded `web_search` reference in the existing fallback-checklist line, and added a renamed-tools test using the child-process config isolation pattern.

Daniishkhan:
Fair point — e7873a2 moves the source of truth to registration time: `index.ts` passes the names it actually registered (skipping any tool disabled via `tools`/`webSearch.enabled`) through `ExtractOptions`, so guidance can never reference an unregistered tool or names edited into the config mid-session. When the caller can't supply names, the guidance degrades to generic wording instead of guessing. Same treatment for the checklist's search bullet. Expanded the test file to cover renamed tools, disabled tools, and the no-names path.

Issue #312, opened 2026-08-26
feat: XCrawl search provider support


zeroicey:
### Platform

- **Name:** XCrawl
- **Website:** https://www.xcrawl.com
- **API docs:** https://docs.xcrawl.com/doc/api-reference/search/
- **What it offers:** SERP search (Google & other engines), plus Scrape / Crawl / Map endpoints. Credits-based billing with a free starter allocation (1,000 credits, no credit card required).

### Why it would be a good fit

XCrawl has been building out an agent ecosystem (official MCP server, n8n nodes, and a skills repo at [xcrawl-api/xcrawl-skills](https://github.com/xcrawl-api/xcrawl-skills)), so there is real demand for first-class agent integrations. The Search API is small, well-documented, and maps cleanly onto the existing provider interface.

### API shape (verified against the live API)

```
POST https://run.xcrawl.com/v1/search
Authorization: Bearer <api_key>
Content-Type: application/json

{ "query": "...", "location": "US", "language": "en", "limit": 10 }
```

Response envelope:

```json
{
  "search_id": "01KKE8BNMEKRHJB9GEWXPYQ8E1",
  "status": "completed",
  "data": {
    "status": "success",
    "credits_used": 2,
    "data": [
      { "position": 1, "title": null, "url": "https://example.com/", "description": "..." }
    ]
  },
  "total_credits_used": 2
}
```

Notes: `limit` is 1–100 (default 10), `location`/`language` are optional, and result items can have `title: null`. Search runs as a server-side job that can take tens of seconds, so it needs a generous client timeout. Because credits are spent per request, it should be explicit-only (like AnySearch/Bright Data/Serper) and excluded from `provider: "all"` and auto fallback.

### Would you be willing to work on this?

Yes — I have a working implementation ready (provider + tests + docs, following the AnySearch/Valyu conventions in this repo). Happy to open a PR if you're open to adding the provider.

nicobailon:
Implemented in PR #313. Thanks again for the XCrawl provider contribution, @zeroicey.

zeroicey:
Thanks for the quick review and merge! Happy to contribute more to this project in the future.

PR #313, opened 2026-08-26
feat: add XCrawl search provider

zeroicey:
### What does this PR do?

Adds [XCrawl](https://www.xcrawl.com) as a new explicit-only search provider, following up on #312.

XCrawl offers SERP search (Google and other engines) via `POST https://run.xcrawl.com/v1/search` with Bearer-token auth. The API shape was verified against the live API; search runs as a server-side job that can take tens of seconds, so the provider uses a 90s timeout.

Since XCrawl spends credits per request, the provider is explicit-only: it is never selected by zero-config `auto`, never included in `provider: "all"`, but can be selected with `provider: "xcrawl"`, configured as the named default, or placed in `searchRouting` — the same conventions as AnySearch / Valyu / Bright Data.

### Implementation notes

- **`xcrawl.ts`** — new provider module modeled on `anysearch.ts`/`valyu.ts`:
  - `xcrawlApiKey` config field + `XCRAWL_API_KEY` env var via the shared `resolveCredential` source chain
  - request body `{ query, limit }` plus optional `location` / `language`; `numResults` clamped to the documented 1–100 range
  - strict envelope validation (`status === "completed"`, `data.status === "success"` when present, `data.data[]` items require non-empty `url`, accept `title: null`)
  - null titles fall back to the result URL so every result keeps a usable display title
  - credentials redacted from error messages and activity logs on both fetch-throw and non-2xx paths
  - availability gated by `hasCredentialSource(...)` (no anonymous tier), matching Valyu/Bocha
- **Registration** — dispatch/availability/label in `gemini-search.ts`; availability map, tool descriptions, schema enums, and explicit-only exclusions in `index.ts`; curator picker option + tag color in `curator-page.ts`/`curator-server.ts`
- **Docs** — new README section, parameter/credential tables, env var lists, file table entry, CHANGELOG entry under Unreleased

### Testing

- `npm run typecheck` passes
- New `test/xcrawl-provider.test.mjs` (5 cases): Bearer auth capture, exact request body, null-title fallback, explicit routing, `limit` clamping to 100, optional `location`/`language` forwarding, HTTP error surfacing without key leakage, malformed-envelope rejection, and unavailable-without-credentials + never-contacted-by-auto-fallback regression
- Full suite: no new failures (the two pre-existing failures in `chrome-cookie-extraction.test.mjs` / `gemini-web-header-overflow.test.mjs` reproduce on upstream `main` without this change)


zeroicey:
Pushed 7850c74 which moves the provider onto the SERP endpoint and addresses all three review findings.

**Why the endpoint change:** during integration testing, `POST /v1/search` consistently returned `status: "completed"` with zero organic results while still charging credits — reproduced across direct connections and multiple networks (JP/US locales, varied payloads, the docs' own example query), with jobs taking ~50–60s. The documented `POST /v1/serp { engine: "google_search", q }` API returns full organic results in a few seconds at half the credit cost, so the provider now uses it and parses SerpApi-style `organic_results`.

Against the review findings:
1. **Provider timeout stops fallback** — fixed: our own timeout now surfaces as `XCrawl search request timed out after Ns` (retriable/network-classified) instead of an abort-shaped error; genuine caller cancellation is detected via the caller's signal state.
2. **Domain filter is discarded** — fixed: the shared include/exclude `domainFilter` is applied client-side to `organic_results`, since the SERP API has no server-side filter.
3. **Locale options lack public path** — fixed by removal: `location`/`language` were dropped along with the old endpoint; the request body is now minimal (`engine` + `q`).

Tests updated: SERP envelope fixtures, client-side domain filter cases (include/exclude), timeout classification regression, plus the existing credential/routing/null-title/malformed-envelope coverage — 6 passing, typecheck clean, no new failures in the full suite.

zeroicey:
Fixed in the latest commit: domainFilter entries now go through the same normalization Valyu applies (trim, lowercase, strip URL form/paths/ports, leading-dash handling, domain-shape validation) before matching, and a leading `www.` is stripped from both sides of the comparison. Added a regression case for `" https://www.Example.COM/path "` style entries alongside the plain include/exclude coverage.

PR #316, opened 2026-08-27
fix: contain Defuddle fallback errors

### LINKED ISSUE #315: v0.25.0 Defuddle fallback logs a raw stack into Pi while returning the fetch as successful
> **Acknowledgement:** This issue was directly observed by the reporter. The complete report, reproduction, and references were created by AI and verified by a human before submission (to the best of his humble abilities). 
> Close right away, if repo has a strict no-AI-assisted-issues-generation policy.

## TL;DR

Some page fetches dump a raw error stack over Pi’s UI, making it unreadable, yet still report success and return the wrong content.

## Summary

During background content fetching from `web_search`, `pi-web-access` 0.25.0 can print a raw `css-select` stack trace directly into Pi’s TUI.

The trigger is React/Next.js streaming markup containing IDs such as `S:a`. Defuddle constructs a selector containing `div#S:a`, where `css-select` interprets `:a` as a pseudo-class.

Defuddle catches the exception internally, logs it through `console.error`, and returns the whole `<body>`. `pi-web-access` then accepts that non-empty fallback as a successful extraction with `error: null`.

The original terminal capture contains private research context, so I cannot attach it. The synthetic reproduction below produces the same stack without personal data or external page content.

## Environment

- macOS 15.7.7
- Pi 0.84.3
- pi-web-access 0.25.0
- Node.js 26.7.0

## Synthetic reproduction

From a `pi-web-access` checkout at `v0.25.0`:

```bash
npm install

node --input-type=module <<'NODE'
import { extractContent } from './extract.ts';

const paragraph = '<p>This is a substantial paragraph of genuine article body text that a content extractor should identify as the main content rather than surrounding navigation or boilerplate.</p>';

const html = `<!doctype html>
<html>
<head><title>Selector reproduction</title></head>
<body>
  <nav><a href="/">Home</a><a href="/about">About</a></nav>
  <template id="B:0"></template>
  <div hidden id="S:a">
    <h1>Selector reproduction</h1>
    ${paragraph.repeat(6)}
  </div>
  <template id="P:a"></template>
  <footer><p>Copyright 2026 Example Inc. All rights reserved.</p></footer>
</body>
</html>`;

globalThis.fetch = async () => new Response(html, {
  status: 200,
  headers: { 'content-type': 'text/html' },
});

const result = await extractContent(
  'https://example.com/repro',
  undefined,
  {
    lookup: async () => [
      { address: '93.184.216.34', family: 4 },
    ],
  },
);

console.log(JSON.stringify({
  error: result.error,
  containsFooter: result.content.includes('Copyright 2026'),
}));
NODE
```

Observed output:

```text
Defuddle Error processing document: Error: Unknown pseudo-class :a
    at compilePseudoSelector (...)
    at compileGeneralSelector (...)
    at rules.reduce._a (.../css-select/lib/esm/compile.js:102:11)
    at compileRules (.../css-select/lib/esm/compile.js:100:18)
    ...

{"error":null,"containsFooter":true}
```

This demonstrates both observable effects:

- The stack is written directly to the process console.
- The extraction is returned as successful while containing page-level footer content.

During an interactive Pi session, the console output becomes interleaved with the active `Working...` or thinking display. During background fetching, the `error: null` result is also counted as a successfully fetched URL.

## Version boundary

The Defuddle fallback was introduced in 0.25.0 by merged PR #306. Version 0.24.2 does not contain this fallback or the Defuddle dependency.

Merged PR #309 intentionally allows thrown Defuddle dependency or runtime failures to remain visible. This case is different: Defuddle catches the selector exception internally, logs it, and returns fallback content, so no exception reaches the `pi-web-access` boundary.

## Relevant code

- Defuddle invocation and returned content:  
  https://github.com/nicobailon/pi-web-access/blob/v0.25.0/extract.ts#L46-L50
- Defuddle output accepted with `error: null`:  
  https://github.com/nicobailon/pi-web-access/blob/v0.25.0/extract.ts#L1228-L1239
- Background fetch success counting:  
  https://github.com/nicobailon/pi-web-access/blob/v0.25.0/index.ts#L1067-L1095

## References

- Defuddle integration, merged PR:  
  https://github.com/nicobailon/pi-web-access/pull/306
- Defuddle error propagation changes, merged PR:  
  https://github.com/nicobailon/pi-web-access/pull/309
- Exact upstream selector bug:  
  https://github.com/kepano/defuddle/issues/352
- Associated upstream PR:  
  https://github.com/kepano/defuddle/pull/358
- Related unsafe class-selector report:  
  https://github.com/kepano/defuddle/issues/361

I found no existing `pi-web-access` issue covering this integration symptom.

The synthetic reproduction and reported version boundary were run and verified before submission.

nicobailon:
Fixes #315.

This contains Defuddle's internal selector-processing console output around the local fallback call. When Defuddle reports a processing error and returns body-level fallback content, `fetch_content` now returns a normal extraction error instead of a successful page body.

Validation:
- `node --test test/rsc-fallback.test.mjs`
- `npm run typecheck`
- `npm test`
- `git diff --check`

Thanks to [@riabiy](https://github.com/riabiy) for the report and reproduction.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@ All notable changes to this project will be documented in this file.
 
 - Cleaned the unreleased XCrawl provider docs, parser shape, provider availability types, and focused tests.
 
+### Fixed
+
+- Kept Defuddle selector-processing failures out of the Pi console and stopped treating the raw page body as a successful extraction. Thanks to [@riabiy](https://github.com/riabiy) for #315.
+
 ## [0.25.0] - 2026-08-25
 
 ### Highlights
--- a/extract.ts
+++ b/extract.ts
@@ -43,10 +43,40 @@ type FetchRouting = { providers: FetchProvider[]; allowRemoteHostedProviders: bo
 const DEFAULT_FETCH_PROVIDER_ORDER: FetchProvider[] = ["http", "firecrawl", "jina", "tinyfish", "search1api", "querit", "kagi", "ollama", "parallel", "brightdata", "gemini"];
 const REMOTE_HOSTED_FETCH_PROVIDERS = new Set<FetchProvider>(["jina", "tinyfish", "search1api", "querit", "kagi", "ollama", "parallel", "parallel-mcp", "brightdata", "gemini"]);
 
+function isDefuddleConsoleError(args: Parameters<typeof console.error>): boolean {
+	const prefix = args[0];
+	return prefix === "Defuddle" || (typeof prefix === "string" && /^Defuddle(?:\s|:)/.test(prefix));
+}
+
 async function extractWithDefuddle(text: string, url: string): Promise<{ title: string; content: string } | null> {
 	const { Defuddle } = await import("defuddle/node");
 	const { document } = parseHTML(text);
-	const result = await Defuddle(document as unknown as Document, url, { markdown: true, useAsync: false });
+	let processingError: unknown;
+	const originalConsoleError = console.error;
+	console.error = (...args) => {
+		if (isDefuddleConsoleError(args)) {
+			if (args[0] === "Defuddle" && args[1] === "Error processing document:") {
+				processingError = args[2];
+			}
+			return;
+		}
+		originalConsoleError(...args);
+	};
+
+	let resultPromise: ReturnType<typeof Defuddle>;
+	try {
+		// With useAsync:false, Defuddle parses synchronously before returning its promise.
+		// Keep the console interception limited to that call so unrelated Pi output is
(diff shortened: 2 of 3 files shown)


riabiy:
Much appreciated and thank you for the effort you put it. 
I respect maintainers and always afraid to use agents for submitting tickets/etc. Yet...have to humble myself and admit, that they produce much better issue reports, than I'd ever be able to in the same amount of time. I'm glad it was helpful, did my best. 

Cheers, D.
(no tokens were burned during this response, only good old-fashioned salt exchange)

PR #318, opened 2026-08-28
fix(xcrawl): resolve API-origin-relative result links to absolute URLs

zeroicey:
## What

`xcrawl.ts` now resolves result `link` values that are not absolute `http(s)` URLs against the XCrawl API origin before emitting them.

## Why

XCrawl's Google SERP (`/v1/serp`, `engine=google_search`) currently returns `organic_results[].link` as API-origin-relative redirect stubs, e.g.:

```json
{ "link": "/goto?url=CAESXAHrOzAV..." }
```

Their docs still show absolute URLs (`"link": "https://openai.com/"`), so this looks like a server-side change. Taking those values verbatim produced broken relative result URLs (and a relative "Source" URL in the answer text). With this change we emit `https://run.xcrawl.com/goto?url=...`, which is a well-formed absolute URL, and non-relative links continue to pass through unchanged.

## Note

This is the defensive provider-side fix. Separately, XCrawl's `/goto` resolver currently returns `403 Access Denied: Only /v1 and /v2 endpoints are available`, so the resulting absolute URL is still an XCrawl-side redirect that only becomes clickable once XCrawl exposes the resolver (or restores absolute links). That upstream regression is worth reporting to XCrawl; see the follow-up for reference.

## Tests

- `test/xcrawl-provider.test.mjs`: added a regression test for relative `/goto` link resolution.
- `npx tsc --noEmit`: clean.
- xcrawl test file: 13/13 pass.
- Full `npm test`: 606 pass / 2 fail — the 2 failures are the pre-existing Gemini cookie/credential environment failures, unrelated to this change.

### DIFF
--- a/test/xcrawl-provider.test.mjs
+++ b/test/xcrawl-provider.test.mjs
@@ -69,6 +69,26 @@ test("XCrawl sends Bearer credentials to the SERP endpoint, normalizes null titl
 	assert.equal(output.routedProvider, "xcrawl");
 });
 
+test("XCrawl resolves API-origin-relative result links to absolute URLs", async () => {
+	const home = await createHome({ xcrawlApiKey: "xc-test-key" });
+	const child = runChild(`
+		globalThis.fetch = async () => new Response(JSON.stringify(${JSON.stringify(serpEnvelope({
+			organic: [
+				{ position: 1, title: "Redirect result", link: "/goto?url=CAESAAAA", snippet: "Obfuscated link." },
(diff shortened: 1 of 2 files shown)


nicobailon:
Thanks again for the XCrawl link fix. I landed the same behavior through owner-branch replacement PR #320 so we could add the required changelog credit and address the whitespace-only link review finding.

The replacement preserves your `/goto?...` relative-link handling and credits your contribution in the changelog. I am closing this PR as superseded by #320.

PR #319, opened 2026-08-28
fix(web_search): expand JSON-array strings in the `query` field

### LINKED ISSUE #317: web_search: JSON-array string in `query` returns 0 results (local models serialize multi-query into the string field)
## Problem

`web_search` intermittently returns **0 results** ("returned 0 sources without answer text", `Successful: 1`) when driven by a local model. Root cause: small/local models serialize multi-query research into a **JSON-array string inside the single-string `query` field**, instead of using the `queries: string[]` parameter. The string is forwarded to the search backend verbatim, so every engine receives a literal `["a","b"]` query and returns nothing.

## Repro

Model tool call captured from a real session (Qwen3.8-27B via a local vLLM, OpenAI-compatible provider, pi 0.84.3, pi-web-access 0.25.0):

```json
{
  "query": "[\"latest stable Android Gradle Plugin version 2026\", \"Jetpack Compose BOM latest stable version 2026\", \"CameraX latest stable version 2026\"]",
  "numResults": 5
}
```

SearXNG side effect (default provider, `searxngBaseUrl` configured): the request hits `/search?q=%5B%22latest+stable+Android...%5D&format=json` — the raw array string as one query — and the engines return zero hits.

Expected behavior: the array is split into N independent queries (each getting its own synthesized answer, like the `queries` parameter does), or the tool returns a validation error the model can correct. Silent 0-results is the worst outcome — the model has no signal that its argument shape was wrong.

## Suggested fix

Normalize `query` before building the query list — if the trimmed string parses as a JSON array of strings, expand it. Patch against `index.ts` (0.25.0), applied at both `rawQueryList` construction sites (the `web_search` execute handler and the `get_search_content` renderCall site):

```ts
// Club-3090 local patch: some local models serialize a list of queries into
// the single-string `query` field as a JSON array (query: "[\"a\", \"b\"]").
// SearXNG then receives the raw blob as one query and returns 0 results.
function expandQueryString(query: unknown): unknown[] {
	if (typeof query !== "string") return [query];
	const trimmed = query.trim();
	if (trimmed.startsWith("[") && trimmed.endsWith("]")) {
		try {
			const parsed: unknown = JSON.parse(trimmed);
			if (Array.isArray(parsed)) {
				const strings = parsed
					.filter((entry): entry is string => typeof entry === "string")
					.map(entry => entry.trim())
					.filter(entry => entry.length > 0);
				if (strings.length > 0) return strings;
			}
		} catch {
			// Not JSON — treat as a literal query string.
		}
	}
	return [query];
}

// execute handler:
const rawQueryList: unknown[] = Array.isArray(params.queries)
	? params.queries.flatMap(expandQueryString)
	: expandQueryString(params.query);
```

Verified locally: with the patch, the exact broken argument shape returns N result groups (one per array element), each with the requested number of sources.

## Context

- pi-web-access 0.25.0, pi 0.84.3, node 24
- Model: Qwen3.8-27B (local vLLM, OpenAI-completions API) — but this will hit any model that isn't strictly schema-literal with the `queries` parameter
- Search backend: local SearXNG instance (JSON format enabled), healthy — direct curl queries return 20-27 results consistently
- Related (already handled well): passing `provider: ["brave","duckduckgo","kagi"]` with no API keys configured produces a clear multi-provider error — no change needed there

Happy to open a PR with the patch if that's easier than applying it from the issue.

zeroicey:
Closes #317.

## Problem

`web_search` returned 0 results when a local model serialized a multi-query list into the single-string `query` field as a JSON array, e.g. `query: "[\"a\", \"b\"]"`. The raw string was forwarded verbatim to every backend (SearXNG receives `q=["a","b"]`), so every engine returned zero hits with no signal to the model that its argument shape was wrong.

## Fix

Added `expandQueryString()`: if a query string parses as a JSON array of strings, expand it into independent queries; otherwise keep it as a single literal query. Applied at both `rawQueryList` construction sites — the `web_search` execute handler and the `renderCall` display path.

## Verification

- **Reproduced first**: a regression test captured the backend receiving one literal `["AGP version", "Compose BOM", "CameraX"]` instead of three queries.
- `test/web-search-query-expand.test.mjs` added; after the fix it asserts three independent backend queries.
- `npx tsc --noEmit`: clean.
- Full `npm test`: 606 pass / 2 fail — the 2 are the pre-existing Gemini cookie/credential environment failures, unrelated to this change.

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -358,9 +358,11 @@ function expandQueryString(query: unknown): string[] {
 	if (trimmed.startsWith("[") && trimmed.endsWith("]")) {
 		try {
 			const parsed: unknown = JSON.parse(trimmed);
-			if (Array.isArray(parsed)) {
+			// Only expand an unambiguously string-only array. Mixed or non-string
+			// arrays are kept as the literal query so we never silently drop
+			// members or collapse them into an empty search.
+			if (Array.isArray(parsed) && parsed.every((entry): entry is string => typeof entry === "string")) {
 				return parsed
-					.filter((entry): entry is string => typeof entry === "string")
 					.map((entry) => entry.trim())
 					.filter((entry) => entry.length > 0);
 			}
@@ -1774,7 +1776,7 @@ export default function (pi: ExtensionAPI) {
 		async execute(callId, params, signal, onUpdate, ctx) {
 			return runWithProxy(typeof params.proxy === "string" ? params.proxy : undefined, async () => {
 				const rawQueryList: unknown[] = Array.isArray(params.queries)
-					? params.queries.flatMap(expandQueryString)
+					? params.queries
 					: (params.query !== undefined ? expandQueryString(params.query) : []);
 				const queryList = normalizeQueryList(rawQueryList);
 				const configWorkflow = loadConfigForExtensionInit().workflow;
@@ -2070,7 +2072,7 @@ export default function (pi: ExtensionAPI) {
 		renderCall(args, theme) {
 			const input = args as { query?: unknown; queries?: unknown };
 			const rawQueryList: unknown[] = Array.isArray(input.queries)
-				? input.queries.flatMap(expandQueryString)
+				? input.queries
 				: (input.query !== undefined ? expandQueryString(input.query) : []);
 			const queryList = normalizeQueryList(rawQueryList);
 			if (queryList.length === 0) {
--- a/index.ts
+++ b/index.ts
@@ -346,6 +346,31 @@ function normalizeQueryList(queryList: unknown[]): string[] {
 	return normalized;
 }
 
+// Some local models serialize a multi-query list into the single-string `query`
+// field as a JSON array (query: "[\"a\", \"b\"]") instead of using the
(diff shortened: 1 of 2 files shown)


zeroicey:
Addressed both review points in commit f8caa41:

1. **Plural queries reinterpreted** — expansion is now scoped to the malformed singular `query` field only; the already-structured `queries` array passes through unchanged, so a literal JSON-array element in `queries` is preserved (added `web_search does not reinterpret already-structured queries entries`).
2. **Mixed arrays lose queries** — `expandQueryString` now expands only when *every* parsed member is a string (`parsed.every(...)`), so mixed/non-string arrays are kept verbatim instead of silently dropping members (added `web_search keeps mixed JSON arrays in query as a single literal query`).

`npx tsc --noEmit` clean; the three `web-search-query-expand` tests pass.

nicobailon:
Thanks again for the fix here. I landed the same behavior through owner-branch replacement PR #321 so we could add the required changelog credit for both the report and implementation.

The replacement preserves your scoped query-array handling and the no-query behavior for empty arrays. I am closing this PR as superseded by #321.

PR #323, opened 2026-08-28
fix(github): isolate clone workdirs across extension processes

MDGChamomile:
## Summary

- create a private `mkdtemp()` namespace under the configured GitHub clone root for each extension runtime
- keep the existing repo/ref hash destinations and process-local `cloneCache` inside that namespace
- make `clearCloneCache()` remove only the current runtime namespace and recreate it on the next clone

## Reproduction

This fixes a reproducible cross-process race. Two extension processes configured with the same clone root currently derive the same destination for the same repo/ref, while their in-memory caches are independent. A failed clone or cache cleanup in one process can therefore remove the other process's in-flight or completed clone.

The regression tests use two Node processes and a controlled fake `gh` executable. They verify that:

- identical repo/ref clones receive different destinations across processes
- one process's failed-clone cleanup does not remove the other process's destination
- successful clones remain readable in both processes
- one runtime's `clearCloneCache()` does not affect another runtime
- repeated requests in one process still reuse the existing `cloneCache`
- cleanup creates a fresh runtime namespace for the next clone

## Scope

This intentionally does not add filesystem locks, cross-process caching, or age-based cleanup for runtime directories left by abnormal termination.

## Validation

- `npm test -- test/github-extract.test.mjs` — 12/12 pass
- `npm run typecheck` — pass
- `npm test` — 614/615 pass; the unrelated `failed password lookups are retried instead of cached` browser-cookie test also fails on unmodified `main` in this environment (16/17 in the focused baseline file)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Fixed
+
+- Isolated GitHub clone workdirs per extension runtime so cleanup in one process cannot delete another process's clone.
+
 ## [0.26.0] - 2026-08-28
 
 ### Highlights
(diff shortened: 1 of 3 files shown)


nicobailon:
Thanks for the clone-runtime isolation fix. I landed the same behavior in #325 with the required changelog credit for @MDGChamomile, plus the follow-up cleanup for failed runtime initialization. Closing this original PR as superseded by #325.

PR #327, opened 2026-08-28
Add fetch.timeout config knob (seconds) to web-search.json

linuxtextadventurer:
## Problem
   `fetch_content` timeouts are hardcoded: 30s on the direct HTTP path, 20s on the Jina path. Slow hosts, large
 documentation pages, and proxied deployments commonly exceed these, with no way to change them short of patching the
 source.

   ## Change
   Optional `fetch.timeout` (seconds, floats allowed) in `web-search.json`:

   ```json
   { "fetch": { "timeout": 300 } }
   ```

   - Read per call — live reload, no restart
   - Unset → current default (30s), existing setups unchanged
   - Invalid value → explicit error naming the config file, no silent fallback
   - Per-call `timeoutMs` precedence unchanged
   - Both the HTTP and Jina paths honor the same knob (Jina currently has its own 20s constant)

   ## Verification
   Checked locally: knob value honored end-to-end, invalid value rejected with a clear error, unset value falls back to
 today's default.

**Provenance:** drafted by an AI coding agent (pi, running Qwen 3.8 27B) in an interactive session; direction, local
 verification, and this description by the submitter.

   ## Compatibility
   Single file (`extract.ts`), no new dependencies, no public API changes; fully opt-in, defaults unchanged.

### DIFF
--- a/extract.ts
+++ b/extract.ts
@@ -33,6 +33,28 @@ import { sanitizeInlineDataUris } from "./data-uri-sanitize.ts";
 const DEFAULT_TIMEOUT_MS = 30000;
 const CONCURRENT_LIMIT = 3;
 
+// fetch.timeout in web-search.json extends the page fetch budget in seconds, shared by the http and jina fallback paths
+// tool-layer timeoutMs option still wins over the knob, lost on pi update --extensions reinstall
+function loadTimeoutMs(): number {
+	const configPath = WEB_SEARCH_CONFIG_PATH;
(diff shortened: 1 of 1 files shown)


nicobailon:
Thanks again for PR #327. I landed the accepted `fetch.timeout` feature through maintainer replacement PR #329 so we could keep your contribution while fixing the current config-error and timer-boundary blockers, plus adding docs, tests, and changelog credit.

Closing this PR as superseded by #329.

PR #328, opened 2026-08-28
Add fetch.answerProvider and fetch.answerModel config knobs to web-search.json

linuxtextadventurer:
## Problem
   `mode: "answer"` always runs on the session's main model. A per-call `answerModel` parameter exists, but there is no
 standing default: a deployment that wants one-shot page QA on a cheap or local model must repeat the model selector on
 every call (or in its system prompt).

   ## Change
   Two optional keys in `web-search.json`:

   ```json
   { "fetch": { "answerProvider": "ollama", "answerModel": "llama3" } }
   ```

   - Both set → default model for answer mode
   - Per-call `answerModel` still overrides the config
   - Neither set → session model (today's behavior)
   - Exactly one set → explicit error: partial config is a mistake, not a guess

   ## Why this helps everyone
   One-shot page QA is a good fit for cheap, fast, or local models — cost savings on bulk extraction, and page content never
 leaving the machine when using local inference. Today that's only possible by re-specifying the model on every call; these
 knobs make it a deployment setting.

   ## Verification
   Checked locally across the full resolution matrix: per-call param beats config, config beats session default, partial
 config errors, unknown model errors.

   **Provenance:** drafted by an AI coding agent (pi, running Qwen 3.8 27B) in an interactive session; direction, local
 verification, and this description by the submitter.

   ## Compatibility
   Single file (`page-query.ts`), no new dependencies, no public API changes; fully opt-in, defaults unchanged.

### DIFF
--- a/extract.ts
+++ b/extract.ts
@@ -33,6 +33,28 @@ import { sanitizeInlineDataUris } from "./data-uri-sanitize.ts";
 const DEFAULT_TIMEOUT_MS = 30000;
 const CONCURRENT_LIMIT = 3;
 
+// fetch.timeout in web-search.json extends the page fetch budget in seconds, shared by the http and jina fallback paths
+// tool-layer timeoutMs option still wins over the knob, lost on pi update --extensions reinstall
+function loadTimeoutMs(): number {
+	const configPath = WEB_SEARCH_CONFIG_PATH;
+	if (!existsSync(configPath)) return DEFAULT_TIMEOUT_MS;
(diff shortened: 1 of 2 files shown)


nicobailon:
Thanks again for PR #328. I landed the accepted answer-default configuration through maintainer replacement PR #330 so we could keep your contribution while fixing the config-precedence and provider-routing blockers, plus adding docs, tests, and changelog credit.

Closing this PR as superseded by #330.

PR #341, opened 2026-09-01
fix(perplexity): keep the citations the answer references

### LINKED ISSUE #340: Perplexity: answer cites sources that numResults truncates away
Ran a search with `provider: "perplexity"`, `numResults: 8`, `workflow: "none"` (v0.27.0). The answer cites [9] through [13], but only 8 sources come back.

Looks like `perplexity.ts`:

```ts
const answer = (data.choices as ...)?.[0]?.message?.content || "";
const citations = Array.isArray(data.citations) ? data.citations : [];

const results: SearchResult[] = [];
for (let i = 0; i < Math.min(citations.length, numResults); i++) {
```

The citations array gets cut to `numResults`; the answer text doesn't. Sonar numbers its markers against the full list, so everything past `numResults` ends up pointing at nothing.

Repro — query `WebMCP best practices`, `numResults: 8`:

```
...multiple guides emphasize selecting only the core, repeatable actions of the product.[1][5][9]
...Prefer lazy/contextual registration so agents only see tools that make sense for the current page state.[2][7]
...Use Chrome-supported debugging or inspector workflows...[1][10]
...Use it when the action depends on the current tab state...[11][13]

**Sources:**
1. Source 1
   https://developer.chrome.com/docs/ai/webmcp/best-practices
... (8 total)
```

Either keep the citations the answer actually references, or renumber the markers to match what survives. Worth considering whether `numResults` should apply to Perplexity at all — for the link-list providers it means "how many results", but for a synthesis provider it's cutting the answer's own bibliography.

schlessera:
Fixes #340.

Perplexity's citations aren't a result list — they're the footnotes for the synthesis. `sonar` numbers its `[n]` markers against the full citations array, so cutting that array to `numResults` leaves the prose pointing at sources that were dropped. `numResults` means "how many results" for the link-list providers; here it's trimming the answer's own bibliography.

This keeps every citation the answer actually references, capped at 20 (matching the existing `numResults` clamp), and lets `numResults` govern only the unreferenced remainder. So a default call still returns 5 unless the answer cites more.

I think this is the right direction, but it's a judgement call on principle that's yours to make — the alternative reading is that `numResults` should cap Perplexity too and the answer text should be renumbered to match. Happy to switch to that, or add a test, if you'd rather go the other way.

Typecheck and `npm test` show no new failures (3 pre-existing `tsc` errors in `content-find.ts` / `openai-search.ts`, and one pre-existing `chrome-cookie-extraction` failure, all present on `main` without this change).

### DIFF
--- a/perplexity.ts
+++ b/perplexity.ts
@@ -105,6 +105,27 @@ export function isPerplexityAvailable(): boolean {
 	});
 }
 
+/** Hard ceiling on kept citations, matching the `numResults` clamp. */
+const MAX_CITATIONS = 20;
+
+/**
+ * How many citations to keep.
+ *
+ * Perplexity's citations are the answer's footnotes, not a result list: sonar
+ * numbers its `[n]` markers against the full array, so cutting the array to
+ * `numResults` leaves the prose referencing sources that were dropped. Keep
+ * every citation the answer actually cites, and let `numResults` govern only
+ * the unreferenced remainder.
+ */
+function citationsToKeep(answer: string, available: number, numResults: number): number {
+	let highestCited = 0;
+	for (const match of answer.matchAll(/\[(\d{1,3})\]/g)) {
+		const index = Number(match[1]);
+		if (Number.isFinite(index) && index > highestCited) highestCited = index;
+	}
+	return Math.min(available, MAX_CITATIONS, Math.max(numResults, highestCited));
+}
+
 export async function searchWithPerplexity(query: string, options: SearchOptions = {}): Promise<SearchResponse> {
 	checkRateLimit();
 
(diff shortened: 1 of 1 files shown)


nicobailon:
Thanks for the report and fix. I landed an owner-branch replacement in #347 so I could add the focused regression coverage and changelog credit while running the maintainer merge gates from a maintainer branch.

The merged change preserves Perplexity sources referenced by numeric citations beyond `numResults`, while keeping the existing `numResults` behavior for answers without citation markers. Credit for [@schlessera](https://github.com/schlessera) is preserved in `CHANGELOG.md`.

PR #343, opened 2026-09-02
fix: scope config-level proxy to extension-initiated requests

mystery4f:
## Problem

`installGlobalProxyFetch()` wraps `globalThis.fetch` process-wide so tool requests can go through the curl transport. But when a request comes from outside any `runWithProxy` scope, `getActiveProxy()` falls back to the config-level `proxy` from `~/.pi/web-search.json`.

Since the wrapper is global, that fallback also captures fetches made by the **host process itself** — e.g. the coding agent's own LLM API calls. With `"proxy": "http://127.0.0.1:4444"` configured, every model request was routed through `curl -x http://127.0.0.1:4444`. When the proxy client is stopped, all of those requests fail with connection errors — even endpoints that are directly reachable and never needed the proxy.

## Fix

Mark extension-initiated async work with an `AsyncLocalStorage` scope:

- `runWithProxy(undefined, fn)` (the "fall back to config-level proxy" path) also marks the context as extension-initiated.
- `getActiveProxy()` returns the config-level `proxy` only inside that scope. Everywhere else it returns `null`, so the global wrapper becomes a transparent pass-through for unrelated callers.

Per-call semantics are unchanged: explicit proxy URL → that proxy; `""` → direct; omitted → config-level proxy, but only for requests made by this extension's own tools.

## Tests

- outside the extension scope, `getActiveProxy()` ignores the config-level proxy; inside it, the proxy applies
- with a config-level proxy set, an unscoped `fetch` delegates to the native fetch (no curl spawn), while a scoped one goes through the curl transport

Also documented in the README that the config-level proxy only applies to this extension's requests.

### DIFF
--- a/README.md
+++ b/README.md
@@ -289,6 +289,8 @@ Raw and direct-image HTTP requests use the same SSRF validation, hostname domain
 
 An empty string (`""`) forces a direct connection even when a config-level proxy is set. Omitting the parameter falls back to the global `proxy` in `~/.pi/web-search.json`.
 
+The config-level `proxy` applies only to requests made by this extension's tools. Other traffic in the host process — such as the coding agent's own model API calls — is never routed through it, so a configured proxy that is temporarily down cannot break unrelated requests.
+
 ```jsonc
 // ~/.pi/web-search.json — global proxy for all tools
 {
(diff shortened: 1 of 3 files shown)


nicobailon:
Thanks for this fix. I landed an owner-branch replacement in #348 that combines the overlapping proxy-scope work from #339 and #343.

The merged change includes this PR's README clarification that config-level proxy applies to extension web-tool traffic rather than host/model requests, and preserves credit for [@mystery4f](https://github.com/mystery4f) in `CHANGELOG.md`.

PR #353, opened 2026-09-04
test: fix three Node-boot races in github-extract / config-path

axelbaumlisto:
Three tests intermittently fail on a loaded host. All three drive a `#!/usr/bin/env node` fake executable and give it a budget shorter than a cold Node boot.

**Found**
- Measured shim boot→first write: 320–940 ms on a busy machine; the tests allowed 500 ms / 1 s.
- Reproduced on pristine `main`: `github-extract` "force-kills…" failed 3 of 4 runs (`ENOENT … processes.json` — process tree SIGKILLed before the shim ran).
- Same class: `github-extract` "disable interactive credential prompts" (`cloneTimeoutSeconds: 1`, fake clone must *finish*) and `config-path` "Gemini command source is lazy…" (`timeoutMs: 1000` bounding two `!command` sh spawns; fetch is mocked).

**Fixed**
- "force-kills": `cloneTimeoutSeconds` 0.5 → 3; the fake `git` now writes both pids **synchronously right after `spawn()`** (the helper pid exists before its runtime boots), so the pid file can never be missing. Still tests the SIGKILL escalation.
- "credential prompts": `cloneTimeoutSeconds` 1 → 4 (outer `spawnSync` budget is 5 s). Not a timeout test.
- "Gemini command source is lazy": `timeoutMs` 1 000 → 15 000.

**Verified**
- Each test ×6 at load >100: 0 failures (before: 0/6, 1/3, 0/3).
- Full suite 716/716 with the patch vs 712/716 without, same host.
- Test-only change; no runtime code touched.

### DIFF
--- a/test/config-path.test.mjs
+++ b/test/config-path.test.mjs
@@ -254,8 +254,10 @@ test("Gemini command source is lazy, overrides stale env, rotates, and uses head
 		const { isGeminiApiAvailable, queryGeminiApiWithVideo } = await import(${JSON.stringify(geminiApiUrl)});
 		const available = isGeminiApiAvailable();
 		const lazy = !existsSync(${JSON.stringify(counterPath)});
-		await queryGeminiApiWithVideo("first", "files/one", { timeoutMs: 1000 });
-		await queryGeminiApiWithVideo("second", "files/two", { timeoutMs: 1000 });
+		// fetch is mocked, so timeoutMs only bounds the two "!command" shell spawns;
+		// keep it generous so a slow CI host cannot abort credential resolution.
+		await queryGeminiApiWithVideo("first", "files/one", { timeoutMs: 15_000 });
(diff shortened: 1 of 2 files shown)


axelbaumlisto:
Valid — the credential-prompt test had a 5 s outer budget against a 4 s clone ceiling, i.e. 1 s for child Node boot + fake `gh` probe + fake `git` boot. Same race class this PR fixes, one level up. Outer budget → 15 s in 2e5b37c (test still finishes in ~2 s when healthy; the extra only matters under load). 3/3 repeated runs green.

nicobailon:
Thanks, @axelbaumlisto. Your original commits landed through #355, with linked changelog credit. The only test follow-up extends the timeout-tree outer budget to cover its clone timeout, kill grace, and cold startup. Full suite: 660/660; focused follow-up: 17/17; typecheck and final-head Greptile passed.

PR #354, opened 2026-09-04
fix(web_search): surface the stored responseId so get_search_content is reachable

axelbaumlisto:
**Found**
- `web_search` stores results under `searchId` but returns it only in `details`, which the model never sees. The output text prints no id (with `includeContent: true` it prints the *fetchId*, a different store).
- `get_search_content` has a full `type === "search"` branch (`query` / `queryIndex`) and its not-found message says "use a responseId returned by web_search" — but nothing ever tells the model that id.
- Observed live: the model guessed `responseId: "latest"` → `No stored results`.

**Fixed**
- Append one line to `web_search` output, mirroring what `fetch_content` already does:
  `Results stored as responseId "<id>". Use get_search_content({ responseId: "<id>", queryIndex: 0 }) to retrieve them.`
- Honours `toolNames.getSearchContent`; omitted when that tool is disabled or when an approved curator summary replaces the raw output.

**Verified**
- New `test/web-search-response-id.test.mjs` (3): printed id === stored `searchId` and retrieval with it succeeds; renamed tool is respected; hint omitted when disabled.
- All 12 existing test files touching `web_search` + `get-search-content`: 89/89.
- `tsc` clean. CHANGELOG entry under Unreleased → Fixed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ All notable changes to this project will be documented in this file.
 
 ### Fixed
 
+- Surfaced the stored `responseId` in `web_search` output so the model can call `get_search_content` on search results; previously the id lived only in `details`, which the model never sees.
 - Retained citations referenced in Perplexity answers even when they fall beyond `numResults`, while preserving the result-count limit for answers without citation markers. Thanks to [@schlessera](https://github.com/schlessera) for issue #340 and PR #341.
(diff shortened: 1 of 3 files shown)


axelbaumlisto:
Valid — the per-run config dir was never removed. Wrapped in try/finally with `rmSync(dir, { recursive: true, force: true })` in 3abd977 so it is cleaned even when an assertion fails. Verified: temp-dir count unchanged across a run (3/3 tests green).

nicobailon:
Thanks, @axelbaumlisto. Your original commits landed through #356, with linked changelog credit. The response-ID implementation and tests are unchanged. The combined current-main tree passed all 663 tests and typecheck; exact-head Greptile passed with no findings.

PR #359, opened 2026-09-05
test(github-issue-pr): assert abort ordering, not wall-clock; shell shims for timed tests

axelbaumlisto:
**Found**
- `github-issue-pr` "REST fallback preserves caller cancellation" fails intermittently: `elapsed` = 1 019 / 1 022 / 1 806 / 6 300 ms against a `< 1000 ms` bound (1 in 4 runs at moderate load, 3 in 4 under heavier load).
- Measured cause: `checkGhAvailable()` spawns the fake `gh` **before** the REST fetch the test times, and process spawn alone is 300–900 ms on a loaded host (even a `/bin/sh` shim; a Node shim adds its own runtime boot on top). `extractContent`'s abort path itself measures 55–66 ms. The bound was racing spawn latency, not our code.
- Two sibling tests ("aborts a hung gh availability probe", "gh timeout") have the same `< 1000` bound around a `gh` spawn.

**Fixed**
- REST test: assert the **ordering invariant** instead of time — the aborted REST request must be the only fetch (`fetched` deep-equals `["https://api.github.com/repos/owner/repo/issues/12"]`), i.e. the generic github.com page fallback never starts. That is what the test is actually protecting.
- The two gh-timeout tests keep a bound (the alternative outcome is a 30 s hang) but at 5 s — equally conclusive, does not race spawn.
- Their `gh` shims are `/bin/sh` scripts (`exit 1`, `sleep 30`, version-then-hang) instead of `#!/usr/bin/env node`: no test value in the Node boot.

**Verified**
- Patched: 16/16 × 3 runs on this branch (`npm ci` deps), plus 0 failures in 8 runs on our fork at load 20–80.
- Control (unpatched file, same host, same session): 1 of 4 runs fails with `got 1019ms`.
- Only `test/github-issue-pr.test.mjs` changes; no runtime code.

### DIFF
--- a/test/github-issue-pr.test.mjs
+++ b/test/github-issue-pr.test.mjs
@@ -15,6 +15,15 @@ async function writeFakeExecutable(binDir, name, source) {
 	return executable;
 }
 
+// For tests that assert a wall-clock bound: a Node shim costs 300–900 ms just
+// to boot on a loaded host and the bound is 1 s, so the probe alone can fail
+// the assertion. A /bin/sh shim starts in a few ms.
+async function writeFakeShellExecutable(binDir, name, script) {
+	const executable = join(binDir, name);
+	await writeFile(executable, `#!/bin/sh\n${script}\n`, { mode: 0o755 });
+	return executable;
+}
+
 test("parseGitHubIssuePrUrl accepts PR and issue variants without changing repo parsing", async () => {
 	const { parseGitHubIssuePrUrl } = await import(moduleUrl);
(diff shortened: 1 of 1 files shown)


nicobailon:
Thanks to [@axelbaumlisto](https://github.com/axelbaumlisto) for the load measurements and focused cancellation-test fix. Fresh review found no concrete blockers. At `9c43121a79c2ce3e282d7c5aab104b19adce47c0`, the focused tests passed 16/16, the full suite passed 661/661, typecheck passed, and Greptile is clear.

Merging the original commit unchanged to preserve authorship. This is test-only; runtime behavior is unchanged.

PR #364, opened 2026-09-07
feat: support 1Password service account credential resolver

Avg8888:
## What Problem This Solves

Pi Web Access credential commands run with a minimized environment that forwards shell-local `OP_SESSION_*` values but omits `OP_SERVICE_ACCOUNT_TOKEN`. As a result, `!op read ...` can reuse an interactive 1Password session but cannot perform unattended reads through a service account already inherited by Pi.

## Why This Change Was Made

Forward `OP_SERVICE_ACCOUNT_TOKEN`, when present in Pi's environment, through the existing credential-command allowlist. The default remains a minimized environment: unrelated variables are still excluded, resolution stays request-time and bounded, and no token is stored in configuration.

The README also makes the authority boundary explicit. An `op://` argument selects what the command requests; it does not narrow the token. Every trusted resolver command receiving the token can exercise the service account's authorized vault permissions, so operators should use narrowly scoped service accounts and trusted absolute executable paths.

## User Impact

Operators can configure provider credentials with a command such as:

```json
{
  "braveApiKey": "!/absolute/path/to/op read 'op://Automation/Brave/credential'"
}
```

and run Pi unattended with an inherited 1Password service-account token. Existing environment aliases, literal credentials, interactive 1Password sessions, command limits, and redaction behavior remain unchanged.

## Evidence

Validated with Node 24.17.0:

- `node --test test/credential-source.test.mjs test/provider-credential-source.test.mjs` — 13 passed
- `npm test` — 665 passed
- `npm run typecheck` — passed
- `npm run audit:runtime` — 0 vulnerabilities
- `git diff --check` — passed

The request-boundary test uses an isolated config, synthetic token, and fake `op` executable. It exercises provider config → shell resolver → minimized child environment → mocked Brave request, verifies unrelated environment variables remain excluded, and confirms the service-account token is absent from the outgoing request URL, headers, and body. No live credentials or provider requests are used.


nicobailon:
Thanks for the contribution. I landed the behavior through the credited maintainer replacement #371, merged as `a879921d7b60fc881f07f577fde7ce664ba0e089`. The replacement preserves the original implementation and commit authorship, and the Unreleased changelog credits both [@Avg8888](https://github.com/Avg8888) and [@xapids](https://github.com/xapids). Closing this superseded PR now that the replacement and post-merge main checks are green.

PR #365, opened 2026-09-08
feat: support SOCKS proxy schemes

phillipzink:
The `proxy` config and per-call `proxy` parameter are validated to only allow `http:`/`https:`, but requests are actually transported via `curl -x <proxy>`, which natively supports SOCKS. This adds `socks4:`/`socks4a:`/`socks5:`/`socks5h:` so a local Tor instance (`"proxy": "socks5h://127.0.0.1:9050"`) or any SOCKS proxy can be used. Additive, non-breaking; verified the proxy travels through curl and routes via Tor's SOCKS5 listener.

### DIFF
--- a/fetch-params.ts
+++ b/fetch-params.ts
@@ -96,7 +96,7 @@ function normalizeAuth(value: unknown): true | string | undefined {
 
 function normalizeProxy(value: unknown): string | undefined {
 	if (value === undefined || value === false) return undefined;
-	if (value === null) throw new Error("proxy must be an http(s) proxy URL string");
+	if (value === null) throw new Error("proxy must be an http(s) or socks proxy URL string");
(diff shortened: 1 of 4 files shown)


nicobailon:
Thanks for the contribution. I landed the SOCKS proxy support through the credited maintainer replacement #372, merged as `ff514f54d6e7afff40ac1f13a1897aa5dd1ddc08`. The replacement preserves the original implementation and authorship, fixes the failing normalization expectation, expands scheme coverage, and credits [@phillipzink](https://github.com/phillipzink) in the Unreleased changelog. The proxy-only DNS suggestion was intentionally not included because explicit curl proxies retain the existing local DNS SSRF preflight. Closing this superseded PR now that post-merge main checks are green.

Issue #280, opened 2026-08-20
Self-hosted Firecrawl instance running on localhost fails with "Firecrawl fallback failed: Blocked internal hostname"


ackalker:
## Description of the problem

When I try to use this extension with a self-hosted Firecrawl instance, I receive the error: "Firecrawl fallback failed", followed by either "Blocked internal hostname: localhost" or "Blocked internal address: 127.0.0.1", depending on whether export `FIRECRAWL_BASE_URL=http://localhost:3002` or `FIRECRAWL_BASE_URL=http://127.0.0.1:3002`.

### Problem

Self-hosted Firecrawl on `localhost` or `127.0.0.1` is blocked by the SSRF guard, and the only documented workaround — adding `127.0.0.0/8` to `ssrf.allowRanges` — is **not documented as a valid use case** for `allowRanges`. The README describes `allowRanges` exclusively as an opt-in for TUN/fake-IP proxies (198.18.0.0/15):

> `ssrf.allowRanges` lists CIDR ranges (e.g. `"198.18.0.0/15"`, `"fd00::/8"`) exempted from the SSRF guard that otherwise blocks private/reserved IP ranges. This unblocks `fetch_content`/`web_search` on hosts whose network proxy runs in TUN + fake-IP mode (Surge, Clash, Mihomo, Stash, ...), where public domains resolve into a synthetic reserved range.

However, the test suite explicitly requires `127.0.0.0/8` in `allowRanges` for self-hosted Firecrawl to work ([firecrawl-extraction.test.mjs](https://github.com/nicobailon/pi-web-access/blob/main/test/firecrawl-extraction.test.mjs#L119-L125), lines 119–125 and 158–163):

```js
test("Firecrawl search honors configured SSRF allow ranges", async () => {
    const home = await configHome({
        firecrawlBaseUrl: "http://127.0.0.1:3002",
        ssrf: { allowRanges: ["127.0.0.0/8"] },
    });
```

### Security concern

Adding `127.0.0.0/8` to `ssrf.allowRanges` does not only exempt the Firecrawl API base URL — it **exempts every URL that resolves to a 127.x.x.x address** from SSRF validation. This includes:

- Any `fetch_content` target whose DNS resolves to 127.x.x.x (e.g. via DNS rebinding)
- Any `web_search` provider that follows redirects to 127.x.x.x
- Any other provider in the extraction fallback chain

[...]

PR #122, opened 2026-07-08
feat: add SearXNG and Firecrawl as self-hosted search/extraction providers

fank:
## Summary

Adds **SearXNG** (self-hosted metasearch) and **Firecrawl** (self-hosted web scraping) as optional providers for `web_search` and `fetch_content`. Addresses the self-hosted portion of [#105](https://github.com/nicobailon/pi-web-access/issues/105).

No breaking changes — these providers are only activated when their environment variables or config keys are set. All existing behavior is preserved.

### SearXNG (`searxng.ts`)
- Lightweight metasearch via SearXNG JSON API (`/search?format=json`)
- Config: `SEARXNG_BASE_URL` env or `searxngBaseUrl` in `web-search.json`
- Supports `numResults`, `recencyFilter`, domain include/exclude filters

### Firecrawl (`firecrawl.ts`)
- **Search** via `/v1/search` endpoint
- **Content extraction** via `/v1/scrape` endpoint (Playwright-rendered Markdown + metadata)
- Auth: `FIRECRAWL_API_KEY` (Bearer token) or `FIRECRAWL_BASIC_AUTH` (HTTP Basic for reverse-proxy setups)
- Config: `FIRECRAWL_BASE_URL` env or `firecrawlBaseUrl` in `web-search.json`

## Integration

- Both providers wired into the **auto fallback chain** in `gemini-search.ts` (local-first priority before SaaS providers)
- Firecrawl added as an **extraction fallback** in `extract.ts` (after Jina Reader, before Parallel/Gemini)
- Both providers exposed in the **curator UI** (provider buttons, labels, CSS tags, server-side validation)
- Provider availability checks (`isSearXNGConfigured()`, `isFirecrawlConfigured()`) skip gracefully when unconfigured

## Implementation Notes

- Follows the same patterns as existing providers (`brave.ts`, `tavily.ts`)
- Firecrawl extraction returns `null` on any failure (not throw), making it a proper fallback tier
- Domain filters applied client-side for consistency across all providers
- Config caching follows the same approach as all other providers (no invalidation until restart)

Refs #105

PR #375, opened 2026-09-10
feat: add self-hosted Crawl4AI extraction fallback

bergheim:
## Problem

`fetch_content` can fall back to a self-hosted Firecrawl instance, but not to [Crawl4AI](https://github.com/unclecode/crawl4ai), the other common self-hosted renderer (single container, Apache-2.0, `POST /md`). Users running Crawl4AI have to leave the tool and call it by hand.

## Fix

A `crawl4ai` fetch provider mirroring the self-hosted Firecrawl extraction path from #165 / #123:

- `crawl4aiBaseUrl` / `CRAWL4AI_BASE_URL` and `crawl4aiApiToken` / `CRAWL4AI_API_TOKEN` (credential-source syntax). Crawl4AI 0.9+ requires a bearer token by default.
- `POST /md` with `{"url", "f": "fit"}`; uses the flat top-level `markdown` only when `success` is `true`; title from the first `# ` heading.
- Runs after `firecrawl` and before the hosted providers, accepted in `fetchRouting.providers`, and treated as self-hosted (not in the remote-hosted set), same as Firecrawl.
- Target URL, API base URL and redirects go through the existing SSRF guard: the `/md` call uses the shared `fetchRemoteUrl` (per-hop validation and hop cap), replays the full POST on every redirect hop, and sends the bearer token only to the configured origin; loopback base URLs are allowed like Firecrawl (#280).
- Crawl4AI has no search API, so `web_search` is untouched.

## Verification

- `test/crawl4ai-extraction.test.mjs` (12 tests): request shape, bearer and credential-source resolution, chain order after Firecrawl, guidance errors, envelope validation, SSRF and redirect behaviour. Each test child gets its own empty config home so a developer's real `web-search.json` cannot leak in, and the other `fetch_content` tests scrub `CRAWL4AI_*` alongside `FIRECRAWL_*`.
- `npx tsc` clean. Full suite: 732 tests, 731 pass; the one failure (`failed password lookups are retried instead of cached`) fails identically on `main` in my environment.
- Live against a self-hosted Crawl4AI 0.9.3: the provider directly and `extractContent` with `fetchRouting.providers: ["crawl4ai"]` both return `# Example Domain` for https://example.com; without a token the 401 surfaces as `Crawl4AI fallback failed: Crawl4AI md error 401`.

Related: #105, #122, #123, #165 (self-hosted Firecrawl), #280 (loopback base URLs).


nicobailon (dismissed):
Thanks for the update and for reusing `fetchRemoteUrl`. One security invariant still blocks merge on this head:

`crawl4ai.ts` passes `allowLoopback: true` into `fetchRemoteUrl` when the configured API base is loopback. `fetchRemoteUrl` reuses that option while validating every redirect, so the exception is not limited to the configured origin.

I reproduced this on `c23d366b4f0d008e0d181723e4cc85415e91394d` with:

- configured base: `http://127.0.0.1:11235`
- `/md` response: `302 Location: http://127.0.0.2:9999/private`
- observed result: both URLs were fetched and the redirected response was accepted

Please restrict loopback redirects to the original configured API origin while preserving POST replay and cross-origin credential stripping, and add a regression for a loopback base redirecting to a different `127.x` or `localhost` origin.

The changelog entry also needs visible contributor credit consistent with the adjacent entries, for example a link to [@bergheim](https://github.com/bergheim) and PR #375.

Current focused tests and typecheck pass, but the redirect case above is not covered.

bergheim:
Confirmed and fixed in a5ded23. Your repro reproduced exactly as written for me too: both URLs fetched, redirected body accepted.

**Where the fix went.** In `fetchRemoteUrl`, not in the provider. The loop reused one options object for every hop, so `allowLoopback` outlived the origin it was granted for. It now records the origin validated on the first hop and drops the exemption as soon as a redirect leaves it. That matches what the option's own doc comment already promises ("explicit provider base endpoints, not fetched targets"), so this reads as a bug in the shared guard rather than a provider special case. No other caller passes `allowLoopback`, so nothing outside the self-hosted paths changes behaviour. Happy to move it into `crawl4ai.ts` instead if you would rather keep `ssrf-protection.ts` out of a provider PR.

Same-origin loopback redirects still work and still replay the POST; cross-origin credential stripping is unchanged.

**Origin of the bug.** It is not new in the rebase or in my `fetchRemoteUrl` refactor. The first version of this PR had it in its own hand-rolled redirect loop, because that loop was copied from `firecrawl.ts`, which computes `allowLoopback` once from the base URL and passes it to every per-hop `validateRemoteUrl`. So `firecrawl.ts` on `main` has the same pivot today, and `brightdata-unlocker.ts` carries a third copy of that loop. I have deliberately left both alone as out of scope here. Firecrawl would be fixed by routing it through `fetchRemoteUrl`, the same change made here, if you want that as a follow-up.

**Regressions.** Three, all failing before the fix and passing after:

- `ssrf-protection.test.mjs`: `allowLoopback` covers a same-origin hop, and does not cover a pivot to `127.0.0.2` or to `localhost`.
- `crawl4ai-extraction.test.mjs`: a loopback base following a same-origin hop then attempting `127.0.0.2` is blocked with only the permitted URLs fetched, and a loopback base redirecting to `localhost` is blocked.

**Changelog** now credits the contributor and links PR #375, matching the neighbouring entries.

Full suite 740/741; the one failure (`failed password lookups are retried instead of cached`) fails identically on `main` in my environment. `npx tsc` clean.

bergheim:
Good catch, fixed in the head commit. The replay now covers only 301 and 302, the hops `fetchRemoteUrl` downgrades and the ones a reverse proxy in front of a self-hosted instance actually emits. 303 keeps the prepared bodyless GET, since it does mean "GET the other resource", and 307/308 already preserve the method untouched.

Covered by a new test asserting that a 303 reaches the destination as a GET with no body while a 307 arrives as the original POST.

bergheim:
Fixed in the head commit. The 303 downgrade is now latched: once a hop has turned the chain into a GET, no later 301 or 302 replays the POST, so the final target cannot receive an extraction body it never asked for. The replay still covers a plain 301/302 chain, which is the case a reverse proxy in front of a self-hosted instance actually produces.

New test walks 303 then 301 and asserts the first hop is the POST and both later hops are bodyless GETs.

nicobailon (dismissed):
The loopback and multi-hop redirect fixes now pass the exact-head tests and Greptile review. One final credential-redaction issue surfaced in independent review:

`response.json()` includes malformed response text in its `SyntaxError`. The catch at `crawl4ai.ts:176` forwards that parser message without applying the existing token redaction, so a malformed Crawl4AI response that contains or reflects a short configured token can expose it through the surfaced error and activity log. This reproduces with a response body equal to a short test token.

Please make the invalid-JSON error generic (preferred), or redact the configured token before including parser details, and add a regression proving the token is absent from the thrown error. The existing non-OK and unsuccessful-envelope paths already redact credentials.

Exact head reviewed: `f39a2fdde2bfef232013f17466bf4dfa4d2ab212`.

bergheim:
Reproduced and fixed in the head commit. With a response body of `c4a-secret` the thrown message was `Crawl4AI md returned invalid JSON: Unexpected token 'c', "c4a-secret" is not valid JSON`, so the token reached both the error and the activity log.

Taken the generic option you preferred: the invalid-JSON path now throws `Crawl4AI md returned invalid JSON` with no parser detail. The status and endpoint already identify what failed, and dropping the detail means there is no body text to redact rather than a redaction that has to stay correct. The non-OK and unsuccessful-envelope paths keep `redactCredential` as before.

New regression sets a short token, returns that exact string as the body, and asserts the thrown message is the generic one and that neither it nor `activityMonitor.getEntries()` contains the token. It fails on the previous commit with the token visible.

Same note as before on where this came from: `firecrawl.ts` builds its invalid-JSON error the same way (`Firecrawl ${label} returned invalid JSON: ${errorMessage(err)}`) and is reachable with a configured token, so the same leak exists there on `main`. Left out of scope here; happy to send it as a separate PR if useful.

Full suite 743/744, the one failure (`failed password lookups are retried instead of cached`) fails identically on `main` in my environment. `npx tsc` clean.

PR #381, opened 2026-09-10
fix: send opencode session headers on the page-answer path

MrSerious0:
## What

`fetch_content` with `mode: "answer"` fails whenever the answer model resolves to opencode:

```
400 {"type":"MissingSessionID","message":"Error from provider (Console Go):
Request is missing x-opencode-session and cannot be routed efficiently."}
```

This includes the default case: with no `fetch.answerProvider` / `fetch.answerModel` pair configured, `answerFromPage()` falls back to `ctx.model`, so any opencode session hits this on every page-answer call.

## Why

`answerFromPage()` dispatches through `ctx.modelRegistry.complete()`, which goes `ModelRegistry.complete` → `ModelRuntime.complete` → `prepareRequest`. That path merges the provider's resolved auth headers plus per-call options, but not Pi's provider attribution headers — `mergeProviderAttributionHeaders()` is called in exactly one place, inside the main agent loop (pi core `dist/core/sdk.js`, `streamFn`). So the main conversation works while this side-call reaches opencode.ai without the required header.

## Fix

Supply the headers from the extension on both dispatch paths:

- registry path → `transformHeaders` (applied by `prepareRequest` immediately before dispatch)
- direct `complete` fallback → `headers`

The session id comes from `ctx.sessionManager.getSessionId()`, deliberately **not** `process.env.PI_SESSION_ID`: that variable is not guaranteed to be present in Pi's own environment when provider auth is resolved, and an unresolved `models.json` header value aborts startup with `Failed to resolve provider "opencode-go" header "x-opencode-session"`.

Headers are added only for opencode models (`provider` is `opencode` / `opencode-go`, or the `baseUrl` host is `opencode.ai`). Other providers are untouched, and an absent session id adds nothing rather than an empty header.

## Tests

Four cases added to `test/page-query.test.mjs`:

- the registry path injects both headers and preserves other request headers
- detection also covers a custom provider routed to an `opencode.ai` base URL
- non-opencode models get no `transformHeaders`
- no session id available → no header

`node --test test/page-query.test.mjs` → 13/13 pass; `npm run typecheck` clean. The full suite shows the same 28 failures before and after the change (Windows shell / curl / `gh` / credential-dependent tests) — verified by running it both with and without the patch; the failure set is identical.

## Note

The underlying gap is arguably in Pi core rather than this extension: any extension calling `modelRegistry.complete()` loses the attribution headers. Happy to rework this or move it upstream if you would rather handle it there.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Fixed
+
+- Send the `x-opencode-session` / `x-opencode-client` attribution headers when `fetch_content` answer mode uses an `opencode` or `opencode-go` model. The answer path dispatches through the model registry, which bypasses the attribution headers Pi merges in the main agent loop, so those requests were rejected with `400 MissingSessionID`.
+
 ## [0.29.0] - 2026-09-10
 
 ### Highlights
--- a/page-query.ts
+++ b/page-query.ts
@@ -98,6 +98,31 @@ function responseText(content: unknown): string {
 	}).join("\n").trim();
 }
 
+/**
+ * pi merges its provider attribution headers (`x-opencode-session` / `x-opencode-client`)
+ * inside the main agent loop only. The registry `complete()` path used below dispatches
+ * straight to the provider, so an opencode model is rejected there with
+ * `400 MissingSessionID` unless the header is supplied by the caller.
+ *
+ * The session id comes from the session manager rather than the environment:
+ * PI_SESSION_ID is not guaranteed to be present in pi's own process.env.
+ */
+function openCodeSessionHeaders(model: Model<Api>, ctx: ExtensionContext): Record<string, string> | undefined {
+	const isOpenCode = model.provider === "opencode"
+		|| model.provider === "opencode-go"
+		|| (() => {
+			try {
+				return new URL(String(model.baseUrl ?? "")).hostname === "opencode.ai";
+			} catch {
+				return false;
+			}
+		})();
+	if (!isOpenCode) return undefined;
+	const sessionId = ctx.sessionManager?.getSessionId?.();
+	if (!sessionId) return undefined;
+	return { "x-opencode-session": sessionId, "x-opencode-client": "pi" };
+}
+
 export async function answerFromPage(
 	input: { question: string; pageText: string; sourceUrl: string; model?: string },
 	ctx: ExtensionContext,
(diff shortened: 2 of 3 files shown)


paltian:
Reproduced the 400 MissingSessionID with an `opencode-go` session model on 0.28.0/0.29.0, and independently tested the same fix approach. Answer mode then returns correct grounded answers. 

Note: the same `typeof registry.complete === "function"` dispatch also exists in `query-rewrite.ts:31` and `summary-review.ts:303`, which this PR doesn't touch. Both would hit the identical 400 with opencode-hosted models (query rewriting and curator summary-review). Worth covering in this PR or tracking as a known limitation?

paltian:
Seems to get fixed in Pi itself on next release :-). See earendil-works/pi#9326. Fix here helps for the time being.

Arteiimis:
Correction to the "fixed in Pi on next release" note: the pi-ai commit ([`561a2e06`](https://github.com/earendil-works/pi/commit/561a2e066c0742e3d3e33c75d5d32a69b1d89194)) adds the header through [`withOpenCodeSessionHeader()`](https://github.com/earendil-works/pi/blob/main/packages/ai/src/providers/opencode-headers.ts#L19), which only fires when `options.sessionId` is set. On `main`, [`packages/coding-agent/src/core/model-runtime.ts`](https://github.com/earendil-works/pi/blob/main/packages/coding-agent/src/core/model-runtime.ts#L590-L593) still has no session-id or attribution merge — [`prepareRequest()`](https://github.com/earendil-works/pi/blob/main/packages/coding-agent/src/core/model-runtime.ts#L592) merges the resolved auth headers with `options.headers` (plus optional `transformHeaders`), and the only caller that supplies a session id is [`sdk.ts`'s `streamFn`](https://github.com/earendil-works/pi/blob/main/packages/coding-agent/src/core/sdk.ts#L314-L334).

So after that release, extension side-calls via [`modelRegistry.complete()`](https://github.com/earendil-works/pi/blob/main/packages/coding-agent/src/core/model-registry.ts#L124) still go out header-less: this path, [`query-rewrite.ts:35`](https://github.com/nicobailon/pi-web-access/blob/main/query-rewrite.ts#L35), and [`summary-review.ts:387-389`](https://github.com/nicobailon/pi-web-access/blob/main/summary-review.ts#L387-L389). This PR remains the fix for the answer path; #385 covers the other two.

Upstream-facing note: passing `sessionId` through the [`modelRegistry.complete()`](https://github.com/earendil-works/pi/blob/main/packages/coding-agent/src/core/model-registry.ts#L124) options would also work once pi-ai ships, since [`prepareRequest()`](https://github.com/earendil-works/pi/blob/main/packages/coding-agent/src/core/model-runtime.ts#L590-L593) spreads unknown options into the provider options — but not on any released version today, so injecting the headers here is the right call for now.

nicobailon:
Thanks @MrSerious0. Your fix was preserved and landed in #397 with profile-linked changelog credit. The promoted code was mechanically equivalent to this reviewed head, rebased onto current `main`, and passed focused tests, typecheck, fresh semantic review, and exact-head Greptile review.

PR #382, opened 2026-09-11
docs: list Crawl4AI in the provider summaries

bergheim:
## Problem

The README header line and the `package.json` description each enumerate every provider the extension supports. Both still list Firecrawl and self-hosted SearXNG but not Crawl4AI, which #375 added in 0.29.0. The npm description is the first thing a reader sees, so the feature is effectively invisible there.

## Fix

- README header: `self-hosted Crawl4AI extraction`, grouped with self-hosted SearXNG rather than dropped into the search list.
- `package.json` description: `Crawl4AI`, next to Firecrawl.
- Changelog entry under Unreleased.

Deliberately unchanged: the `web_search` provider list in the README, the `provider` parameter table, and the search tool description in `index.ts`. Crawl4AI has no search API, so it does not belong in any of them. The rest of the README already covers it: fallback chain prose, the How It Works diagram, the config example, `fetchRouting.providers`, the credential-source and environment-variable lists, the extraction timeout list, and the file table.

## Verification

Documentation and metadata only, no behaviour change. `npx tsc` clean; full suite 744/745, the one failure (`failed password lookups are retried instead of cached`) fails identically on `main` in my environment. `jq` round-trips `package.json`.

Related: #375.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Fixed
+
+- List Crawl4AI in the README provider summary and the package description, which both still omitted it after the provider shipped in 0.29.0. Thanks to [@bergheim](https://github.com/bergheim) for [PR #382](https://github.com/nicobailon/pi-web-access/pull/382).
+
 ## [0.29.0] - 2026-09-10
 
 ### Highlights
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@
 
 # Pi Web Access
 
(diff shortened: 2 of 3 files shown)


nicobailon:
The documentation change is accurate and the diff is clean. Crawl4AI is described as extraction rather than search, and the existing detailed docs preserve that distinction.

The PR is still marked draft, which I’m treating as the contributor’s hold. Please mark it ready when you want it considered for merge; no technical rework is currently requested.

nicobailon:
Superseded by #410, which reapplied this documentation and package-metadata change on current `main` while preserving subsequent provider and changelog updates.

Thank you, [@bergheim](https://github.com/bergheim), for the original change in #382. The promotion PR and Unreleased changelog retain your profile-linked credit.

PR #384, opened 2026-09-11
fix: prevent uncaughtException crash from abort() in timer/event callbacks

Max19970:
## Problem

A bare `AbortController.abort()` (called with no reason) constructs a `DOMException` via `globalThis.DOMException` to populate `signal.reason`. When the host runtime has installed a throwing/fake `DOMException` — for example the `node-domexception` polyfill that gets bundled into some downstream builds — that construction throws.

When this happens inside a `setTimeout` or a `"close"` event callback, the throw escapes the callback as an **uncaughtException** and crashes the entire host process.

Observed crash in the wild (pi coding agent):

```
DOMException [AbortError]: This operation was aborted
    at new DOMException (node:internal/per_context/domexception:76:18)
    at AbortController.abort (node:internal/abort_controller:572:18)
    at Timeout._onTimeout (.../pi-web-access/summary-review.ts:312:26)
    at listOnTimeout (node:internal/timers:685:17)
```

`summary-review.ts:312` is the summary-generation deadline timer firing `deadlineController.abort()`.

## Fix

Pass an **explicit abort reason** to `abort()` so it never constructs a `DOMException` internally, and wrap the calls in `try/catch` so nothing can ever throw out of a timer or event callback:

- **`summary-review.ts`** — the deadline `setTimeout` callback and `checkSummaryDeadline()` now call `deadlineController.abort(deadlineMarker)` inside `try/catch`.
- **`extract.ts`** — the fetch-timeout timer and the caller-abort listener now pass an explicit reason and are guarded.
- **`curator-server.ts`** — the `req.on("close")` handler now passes an explicit reason and is guarded.

Using an explicit reason is also semantically nicer: `signal.reason` now carries the marker/label instead of a generic `AbortError`, and nothing in the codebase reads `.reason` in a way that depends on it being a `DOMException`.

## Testing

Added `test/abort-domexception-crash.test.mjs`, which spawns a child that:

1. installs a `globalThis.DOMException` that throws when used to build an `AbortError` (mimicking the broken polyfill),
2. spies on `AbortController.prototype.abort` to assert the deadline abort passes an explicit reason, and
3. asserts the process survives without an `uncaughtException`.

The test **fails on the unpatched code** (`NO_EXPLICIT_REASON`) and **passes with the fix** (`REASON_IS_MARKER`).

Note: on Node 24, `abort()` uses an internal `DOMException` rather than the global, so the literal crash only reproduces on affected/older runtimes — the test therefore guards the explicit-reason contract, which is what prevents the crash everywhere.

## Scope

Only `abort()` call sites that run inside timer or event callbacks (where a throw becomes an uncaughtException) were changed. Bare `abort()` calls in normal request/response call stacks are left as-is since a throw there propagates to an awaiting caller rather than crashing the process.

### DIFF
--- a/curator-server.ts
+++ b/curator-server.ts
@@ -584,7 +584,11 @@ export function startCuratorServer(
 					return;
 				}
 				const controller = new AbortController();
-				req.on("close", () => controller.abort());
+				// Guard + explicit reason so abort() never throws inside the "close"
+				// event callback (which would escape as an uncaughtException).
+				req.on("close", () => {
+					try { controller.abort("request-closed"); } catch { /* never throw */ }
+				});
 				touchHeartbeat();
 				try {
 					const rewritten = await callbacks.onRewriteQuery(query.trim(), controller.signal);
--- a/extract.ts
+++ b/extract.ts
@@ -1153,9 +1153,17 @@ async function extractViaHttp(
 
 	const controller = new AbortController();
 	const startedAt = Date.now();
-	const timeoutId = setTimeout(() => controller.abort(), timeoutMs);
+	// Pass an explicit reason and guard so abort() never constructs a
+	// DOMException or throws inside a timer/event callback. On runtimes where
+	// globalThis.DOMException is a throwing polyfill (e.g. node-domexception),
+	// a bare abort() would escape as an uncaughtException and crash the process.
+	const timeoutId = setTimeout(() => {
+		try { controller.abort("timeout"); } catch { /* never throw from timer */ }
+	}, timeoutMs);
 
-	const onAbort = () => controller.abort();
+	const onAbort = () => {
+		try { controller.abort("caller-aborted"); } catch { /* never throw */ }
+	};
 	signal?.addEventListener("abort", onAbort);
 
 	try {
--- a/summary-review.ts
+++ b/summary-review.ts
@@ -309,7 +309,16 @@ export async function generateSummaryDraft(
 	let deadlineTimer: ReturnType<typeof setTimeout> | undefined;
 	const deadlinePromise = new Promise<typeof deadlineMarker>(resolve => {
 		deadlineTimer = setTimeout(() => {
-			deadlineController.abort();
+			// Pass an explicit reason so AbortController.abort() never constructs a
+			// DOMException internally. Some runtimes install a throwing/fake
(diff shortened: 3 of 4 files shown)


nicobailon:
Thanks for tracing the reported timer site. Two current-head blockers need resolution before this can merge:

1. `test/abort-domexception-crash.test.mjs` can return through `summary-model-unavailable` before its 5 ms deadline, so it does not deterministically exercise the abort timer. This reproduces locally as `NO_EXPLICIT_REASON`. Please keep an eligible completion pending beyond the deadline and assert the public timeout result.
2. The `extract.ts` and `curator-server.ts` changes expand beyond the demonstrated summary timer, but the PR has no public timeout/caller-abort/request-close coverage for them. Please either remove those unrelated changes or add deterministic behavior-level coverage.

The claimed global `DOMException` crash mechanism also does not reproduce on the current local Node runtime, where `AbortController.abort()` uses the runtime’s internal DOMException. Please identify the affected supported Node/runtime version and demonstrate base failure/head success there, or narrow the claim and patch to the behavior the test can prove.

This is separate from #388: that report concerns ordinary prompts with no web-tool call and has no stack connecting it to these callbacks.

nicobailon:
Thanks @Max19970 for tracing the stack to the summary deadline and investigating abort reasons.

I reproduced the current PR head on Node v25.2.1 / Pi 0.85.1. The regression test does not reach the intended deadline path: no eligible model is available, so summary generation returns early and the test fails with:

```
NO_EXPLICIT_REASON
SURVIVED true
```

A direct runtime probe also shows that bare `AbortController.abort()` returns synchronously. The `DOMException` stack points to `abort()` because that is where the signal reason is created; a later rejected promise retains that stack. Changing the reason to a `Symbol` or string does not handle that rejection, and the surrounding `try/catch` cannot catch an asynchronous rejection.

The patch therefore lacks a deterministic base-fail/head-pass reproduction and changes standard `AbortError` semantics without proving that those call sites caused the process exit. The two Greptile findings remain valid, so I’m closing this PR rather than merging it.

A future fix should identify the promise consuming the summary completion signal, handle or await its rejection at that boundary, and include a deterministic timeout test with an eligible pending model that proves there is no `unhandledRejection` or `uncaughtException`. We’ll preserve your reporting/investigation credit if that work uses this diagnosis.

PR #386, opened 2026-09-12
feat: add Serply search provider

googio:
Disclosure: I work with Serply, so treat this as a vendor-authored integration and review it accordingly. Happy to adjust scope, naming, or drop it entirely if it is not a direction you want for the project.

## What

Adds `serply.ts`, an explicit-only [Serply](https://serply.io) Google SERP provider, alongside `serpbase.ts`, `serpapi.ts`, and `serper.ts`. The wiring follows the SerpApi provider merged in #374.

## Why

Serply is one more key-based Google SERP API, so this is additive for users who already have a key and costs nothing to everyone else. Like SerpBase, SerpApi, and Serper it stays out of `auto` and out of `provider: "all"`, because each request consumes paid SERP credits. It can still be selected explicitly, configured as the default provider, or placed in `searchRouting`.

## Shape

`serply.ts` is a close mirror of `serpapi.ts`, including the three corrections you made on top of #363:

- a response-body `TimeoutError` is re-thrown from the JSON `catch` instead of being reported as invalid JSON, so a deadline abort still routes as a network failure
- `options.signal?.aborted` is checked before the provider timeout, so caller cancellation surfaces as `Aborted` rather than a timeout
- `redactCredential` runs over the outgoing message and over JSON error envelopes, and the original error identity is preserved when redaction changes nothing

Four differences from `serpapi.ts`, all forced by the vendor API:

| | SerpApi | Serply |
| --- | --- | --- |
| auth | `api_key` query parameter | `X-Api-Key` request header |
| results array | `organic_results` | `results` |
| snippet field | `snippet` | `description` |
| error envelope | `error` | `detail` |

`q`, `num`, and `tbs` are the same query parameters, so domain filtering and `recencyFilter` behave exactly as they do for SerpApi: include/exclude domains are sent as Google `site:` / `-site:` clauses and reapplied locally against the returned hostnames, and `day`/`week`/`month`/`year` map to `qdr:d` / `qdr:w` / `qdr:m` / `qdr:y`.

## Notes

- Serply support is optional; behavior is unchanged when `serplyApiKey` and `SERPLY_API_KEY` are both unset.
- Because the key travels in a header rather than a query parameter, it cannot end up in a logged request URL.
- Docs updated in `README.md` (tagline, Zero Config, `web_search` prose, the `provider` table row, the config sample, both credential-source lists, a per-provider section, the env var list, the `provider: "all"` section, and the file manifest), plus `CHANGELOG.md` and the package description.
- One file #374 did not need: `isAvailableProvider` in `curator-server.ts` is an `if` chain ending in `return false`, so a provider with no branch there is rejected when the curator switches to it. I added the `serply` branch. It looks like `serpapi` is missing the same branch on `main` today. I left that alone to keep this diff focused, but say the word and I will add the one-liner.

## Testing

`test/serply-provider.test.mjs` mirrors `test/serpapi-provider.test.mjs`, 11 cases, all network mocked. It covers result mapping, domain include/exclude, `tbs` recency, the `num` bump when domain filters are active, an assertion that the key is in the header and **not** in the query string, explicit routing, exclusion from `provider: "all"`, timeout fallthrough to a configured next provider, the six delayed-body cancellation cases, credential redaction in both a JSON `detail` envelope and a non-JSON error body, and the Curator entry.

```
$ node --test test/serply-provider.test.mjs
✔ Serply maps Google results, filters, recency, and environment credentials (153.832446ms)
✔ Serply supports explicit routing but is excluded from provider all (193.611962ms)
✔ Serply provider timeouts can fall through configured routing (193.397588ms)
✔ Serply delayed 200 body honors deadline cancellation routing (1209.625508ms)
✔ Serply delayed 200 body honors caller cancellation routing (224.811494ms)
✔ Serply delayed 200 body honors both cancellation routing (1197.195327ms)
✔ Serply delayed 503 body honors deadline cancellation routing (1205.650033ms)
✔ Serply delayed 503 body honors caller cancellation routing (226.332571ms)
✔ Serply delayed 503 body honors both cancellation routing (1203.639828ms)
✔ Serply redacts the resolved credential in JSON error envelopes (132.605827ms)
✔ Serply redacts API errors and appears in the Curator (179.214892ms)
ℹ tests 11
ℹ pass 11
ℹ fail 0
```

`npm run typecheck` is clean.

`npm test` gives 754 pass / 2 fail out of 756. Both failures are pre-existing and unrelated to this change:

```
✖ failed password lookups are retried instead of cached    test/chrome-cookie-extraction.test.mjs
✖ a failed clone cannot clean up another process's in-flight destination    test/github-extract.test.mjs
```

I reproduced exactly those two on a clean worktree of `main` at 192ac18 with this branch removed, so they are environmental here (Linux/WSL2, Node 24.11.1) rather than introduced by the provider.

Live check against the real API with my own key, `numResults: 3`, through the new module rather than a mock:

```
isSerplyAvailable(): true

[1] plain search, numResults 3 -> 3 results
  - Understanding Ownership - The Rust Programming Language | https://doc.rust-lang.org/book/ch04-00-understanding-ownership.html
  - Ownership as explained in the Rust book - Reddit | https://www.reddit.com/r/rust/comments/yibdpi/ownership_as_explained_in_the_rust_book/
  - What is Ownership? - The Rust Programming Language | https://doc.rust-lang.org/book/ch04-01-what-is-ownership.html

[2] domainFilter ['doc.rust-lang.org'] -> 3 results
  - https://doc.rust-lang.org/book/ch04-00-understanding-ownership.html
  - https://doc.rust-lang.org/book/ch04-01-what-is-ownership.html
  - https://doc.rust-lang.org/nomicon/ownership.html

[3] recencyFilter 'day' -> 3 results
  - Awesome Agent Orchestrators Awesome - GitHub | https://github.com/andyrewlee/awesome-agent-orchestrators
  - Paperclip - The app people use to manage AI agents for work | https://paperclip.ing/
  - This Open-Source AI Agent Can Run for Days - YouTube | https://www.youtube.com/watch?v=TMGvXxZWY3o
```

Error shapes, also verified live: a wrong key returns HTTP 401 with `{"detail":"Invalid API key"}` and a missing key returns HTTP 401 with `{"detail":"API key missing"}`. Both are surfaced through the same redaction path as the other providers. API docs are at https://serply.io/docs.


nicobailon:
Thanks @googio. Your Serply provider contribution was preserved and landed through #402 with the original contributor commit retained and profile-linked vendor credit in the changelog. The replacement also fixed cross-origin credential stripping, result-URL validation, cancellation/redaction coverage, and complete provider documentation.

PR #390, opened 2026-09-14
fix: ground PDF answers in extracted Markdown

MDGChamomile:
## Problem

PDF extraction succeeds, but `fetch_content` with `mode: "answer"` supplies only the saved-file notice (path, pages, and character count) to `answerFromPage`, not the PDF text.

## Fix

- Return the already-generated Markdown alongside PDF extraction metadata.
- Use it as fetched content only in answer mode; readable-mode file output and file saving remain unchanged.
- Reuse the existing answer-input limits, sanitization, and original-content retrieval without rereading a shared output filename.

No dependency, configuration, or model-routing changes.

## Validation

Base: `192ac1875e3b8f88c78953dbc314949ec9fcaa27` (v0.29.0), Node 22.22.3, TypeScript 7.0.2, unchanged lockfile.

- New regression tests on the unmodified base: 4 pass / 3 fail, reproducing the missing body input.
- All 7 new regression tests pass with this change. They generate real PDFs locally, mock HTTP and model calls, and check body delivery, saved files, retrieval, concurrent filenames, truncation, HTML compatibility, disabled extraction, raw rejection, and cancellation.
- Existing PDF engine tests additionally compare returned Markdown with the saved file.
- Typecheck passes.
- Full suite including packed-install coverage: **751 pass / 1 fail**; base: **744 pass / 1 fail**.
- The same unrelated `failed password lookups are retried instead of cached` test fails on both revisions on Linux. The suite is not claimed to be fully green.
- No live AI service calls; npm lifecycle scripts disabled during temporary dependency installation.

```sh
NODE_OPTIONS=--experimental-strip-types node --test test/fetch-pdf-answer.test.mjs
npm run typecheck
NODE_OPTIONS=--experimental-strip-types node --test --test-concurrency=3
```

The pre-existing Linux test failure is addressed separately in #389. This PR is based directly on main and does not include that test change.


nicobailon:
Thanks @MDGChamomile. Your PDF grounding fix was preserved and landed in #398 with profile-linked changelog credit. The promoted change passed focused PDF tests, typecheck, fresh semantic review, and exact-head Greptile review.

PR #391, opened 2026-09-14
fix: return bounded excerpts for overflowing content matches

MDGChamomile:
## Problem

Dense nearby matches merge into an arbitrarily large excerpt. When it exceeds the 20,000-character response budget, formatting stops without returning any excerpt. With sparse matches, frequent early occurrences can also consume the budget before a different query's later occurrence is shown.

Minimal reproduction on main:

```js
const result = findContent("common context.\n\n".repeat(4_000), ["common"], "exact");
// matchCount: 4000, returnedMatches: 0
```

The existing cap test checks the upper bound but does not require any match excerpts to be returned.

## Fix

Keep the existing result unchanged when all excerpts fit. Only on overflow:

- Bound merged ranges to 1,600 source characters instead of growing them indefinitely.
- Prioritize representative ranges across matched queries, then fill remaining space with other occurrences. This is budgeted sampling, not a new relevance-ranking system or a promise to return every occurrence.
- Allow a later smaller excerpt to fit and reserve space for missing-query/truncation notices.

Matching semantics, total/per-query counts, tool parameters, and the complete 20,000-character output limit remain unchanged. No dependencies are added.

## Validation

Base: `192ac1875e3b8f88c78953dbc314949ec9fcaa27` (v0.29.0), Node 22.22.3, TypeScript 7.0.2, unchanged lockfile.

- Regression file before the production fix: **4 pass / 9 fail**.
- After the fix: **13/13 finder tests pass**, plus a tool-level stored-content regression in the full suite.
- Covers dense exact/case-insensitive/fuzzy matches; sparse and dense common-query starvation in both query orders; unchanged fitting-response formatting; whitespace normalization; full notices; and ten maximum-length query labels.
- Typecheck passes.
- Full suite including packed-install coverage: **756 pass / 1 fail**; base: **744 pass / 1 fail**.
- The same unrelated Linux `failed password lookups are retried instead of cached` failure remains; this PR does not modify cookie handling or its tests.

```sh
NODE_OPTIONS=--experimental-strip-types node --test test/content-find.test.mjs test/get-search-content.test.mjs
npm run typecheck
NODE_OPTIONS=--experimental-strip-types node --test --test-concurrency=3
```

The pre-existing Linux test failure is addressed separately in #389. This PR is based directly on main and does not include that test change.


nicobailon:
Thanks @MDGChamomile. Your bounded excerpt contribution was preserved and landed through #399 with the original contributor commit retained and profile-linked changelog credit. The replacement completed adversarial query coverage, disjoint excerpt handling, exact output budgeting, and indexed dense-match scoring.

PR #405, opened 2026-09-14
fix: honor configured fetch provider order

### LINKED ISSUE #393: fetchRouting.providers still runs HTTP extraction first when HTTP is omitted or reordered
## Summary

`fetchRouting.providers` does not fully control provider selection/order: remote URLs get an implicit HTTP extraction even when `http` is omitted or placed after another provider. If that extraction succeeds, the configured provider is never reached.

Observed while investigating `pi-web-access` 0.29.0; the same logic is present in main at `192ac1875e3b8f88c78953dbc314949ec9fcaa27`.

## Background

A product-page price lookup returned useful-looking content from HTTP/readable extraction but omitted the requested price. A separate direct request to the configured Firecrawl API, using the extension's `formats: ["markdown"]` and `onlyMainContent: true` options, returned Markdown containing the price.

This makes temporarily preferring Firecrawl a reasonable workflow, even with its additional cost and latency. However, changing provider routing cannot reliably force that preference.

Relevant starting configuration (credentials and endpoint omitted):

```json
{
  "fetchRouting": {
    "providers": ["http", "firecrawl", "brightdata"],
    "allowRemoteHostedProviders": true
  },
  "firecrawlApiVersion": "v2",
  "firecrawlFreshScrape": true
}
```

## Current behavior

In [`extract.ts`](https://github.com/nicobailon/pi-web-access/blob/192ac1875e3b8f88c78953dbc314949ec9fcaa27/extract.ts#L813-L816):

```ts
if (remoteUrl && providerOrder[0] !== "http") {
  const httpGateResult = await runHttpProvider();
  if (httpGateResult) return httpGateResult;
}
```

`runHttpProvider()` returns successful HTTP extraction as well as certain terminal errors. Consequently, even this configuration can return HTTP-extracted content without calling Firecrawl:

```json
{
  "fetchRouting": {
    "providers": ["firecrawl"],
    "allowRemoteHostedProviders": true
  }
}
```

This conclusion comes from inspecting the routing code, rather than a separate end-to-end run with that modified configuration. The price comparison used a direct Firecrawl API request.

## Requested change

Honor the configured provider selection and order literally. Keep URL/SSRF validation and required redirect-policy enforcement mandatory, but separate those safety checks from selecting a successful content extraction.

If an HTTP safety preflight is necessary, it should not silently supply the final content when another provider was selected. Document any unavoidable preflight network activity separately from extraction routing.

Raw mode currently takes a separate HTTP-only path. That distinction should remain explicit rather than implying that this routing also governs raw requests.

nicobailon:
## Summary

Honors configured `fetchRouting.providers` order instead of forcing direct HTTP extraction before a selected hosted provider.

- removes the duplicate pre-routing HTTP gate
- keeps mandatory initial URL/domain/SSRF validation before all extraction
- preserves raw direct-HTTP behavior, redirect safety, cancellation, bounds, defaults, and remote-provider opt-in
- credits [@Slooz](https://github.com/Slooz)

## Validation

- focused routing/raw/timeout and provider SSRF suites
- full repository tests: 748/748
- TypeScript typecheck
- `git diff --check`
- retained challenges/deslop/simplify and fresh final review: OK

Closes #393.


nicobailon:
Closing this candidate after the final security review reproduced two unresolved P1 contract failures. A local one-byte preflight does not validate the request a hosted provider independently performs: redirects, DNS resolution, request headers, and terminal media type can differ. It therefore cannot guarantee per-hop SSRF/domain validation or enforce disabled PDF/image policy. The preflight also exposed uncancelled redirect bodies.

The branch will not be merged. The existing HTTP-first path remains the enforced safety boundary.

